PR #31, opened 2023-03-24
Fix panic in newExecutor when fetching RenterKey for MDM programs without contract

ChrisSchinnerl:


n8maninger (commented):
Looks good. Nice catch!

PR #79, opened 2023-06-12
Change collateral setting to collateral multiplier

n8maninger:
Changes the collateral setting to a multiplier.


n8maninger:
@alexfreska

Issue #18, opened 2023-03-06
Cache commonly used sectors in memory


n8maninger:
Add a configurable LRU cache for sectors. 

_Edit: add implementation details_

---

The cache needs to link the sector's root (`types.Hash256`) with the sector's data (`*[rhpv2.SectorSize]byte`). Older sectors should be ejected as more sectors are added. The cache should be disabled if the setting is zero. 

There are a couple of moving parts here. My recommendation is to start with a static LRU cache in the volume manager, write some tests, then move on to making it configured via the host's settings. Each part can be together or separate PRs.

### Part 1 - Add an LRU cache to the volume manager
1. Add the cache to the volume manager https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/host/storage/storage.go#L63-L78
2. Check the cache for the sector root. If the sector does not exist in the cache, read it from disk, and add it. https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/host/storage/storage.go#L844-L871
3. Add newly written sectors to the cache https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/host/storage/storage.go#L901-L927 

### Part 2 - Make the cache size configurable
1. Add a new setting for the number of sectors to cache. An `int` or `uint32` is fine. Each sector is 4MiB. https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/host/settings/settings.go#L39-L76
2. Add the new setting to the database https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/persist/sqlite/init.sql#L184-L209
3. Add a migration to add the new column https://github.com/SiaFoundation/hostd/blob/master/persist/sqlite/migrations.go
4. Change the store methods to read and update the new column https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/persist/sqlite/settings.go#L17-L112
5. Pass the initial cache setting into the volume manager and initialize the cache https://github.com/SiaFoundation/hostd/blob/16d5e8efc93fafeddb0b37b6b1e957d089c657ff/host/storage/storage.go#L956-L979
[...]

PR #80, opened 2023-06-13
Add an LRU cache to the volume manager 

SAHU-01:
See issue #18 Part 1:

- [X] Add the cache to the volume manager

- [X] Check the cache for the sector root. If the sector does not exist in the cache, read it from disk, and add it.

- [X] Add newly written sectors to the 

- [ ] Write unit tests


host/storage/storage.go
```diff
@@ -74,6 +76,8 @@ type (
 		volumes map[int]*volume
 		// changedVolumes tracks volumes that need to be fsynced
 		changedVolumes map[int]bool
+		cache          *lru.Cache[types.Hash256, *[rhpv2.SectorSize]byte] // Added cache
+		cacheMu        sync.Mutex
```
n8maninger, line 80: cacheMu should be unnecessary. lru.Cache has its own locking

host/storage/storage.go
```diff
@@ -955,6 +980,8 @@ func (vm *VolumeManager) PruneSectors() (int, error) {
 
 // NewVolumeManager creates a new VolumeManager.
 func NewVolumeManager(vs VolumeStore, a Alerts, cm ChainManager, log *zap.Logger) (*VolumeManager, error) {
+	// Initialize cache with LRU eviction and a max capacity of 1000 items
+	cache, _ := lru.New[types.Hash256, *[rhpv2.SectorSize]byte](1000)
```
n8maninger, line 984: 1000 sectors is 4GiB, the default maximum should be 64 (256MiB of memory).
Once the cache is configurable users with more memory can increase the cache size.

host/storage/storage.go
```diff
@@ -967,12 +994,24 @@ func NewVolumeManager(vs VolumeStore, a Alerts, cm ChainManager, log *zap.Logger
...
+		cacheMu:        sync.Mutex{},
 
 		tg: threadgroup.New(),
 	}
 	if err := vm.loadVolumes(); err != nil {
 		return nil, err
 	}
+
+	// Background goroutine to evict old entries from cache
+	go func() {
```
n8maninger, line 1007: ?

n8maninger (changes requested):
Nice job! Just a couple of comments

n8maninger:
It would also be nice to track cache hits and misses. Adding counters to the volume manager and incrementing them with the `atomic` package should make that very straightforward.

n8maninger:
`go.sia.tech/hostd/rhp/v3/TestStoreSector` is failing due to cache hit. You should be able to fix it with a cache eject `rpc_test.go:220: expected error when reading sector`

host/storage/storage.go
SAHU-01: Yes, you are right.The background goroutine that sleeps for a minute and locks/unlocks the mutex does not actually do anything useful and can be removed.

SAHU-01:
> `go.sia.tech/hostd/rhp/v3/TestStoreSector` is failing due to cache hit. You should be able to fix it with a cache eject `rpc_test.go:220: expected error when reading sector`

In the updated code, after detecting a cache hit, the sector is returned as expected, but before doing so, the sector is removed from the cache using `vm.cache.Remove(root)` to simulate a cache miss. I hope the test should pass after making this change, do let me know your thoughts.

host/storage/storage.go
```diff
@@ -850,6 +857,18 @@ func (vm *VolumeManager) Read(root types.Hash256) (*[rhpv2.SectorSize]byte, erro
 	}
 	defer done()
 
+	vm.mu.Lock()
```
n8maninger, line 860: This introduces a deadlock

host/storage/storage.go
```diff
@@ -850,6 +857,18 @@ func (vm *VolumeManager) Read(root types.Hash256) (*[rhpv2.SectorSize]byte, erro
...
+	defer vm.mu.Unlock()
+
+	// Check the cache first
+	if sector, ok := vm.cache.Get(root); ok {
+		atomic.AddUint64(&vm.cacheHits, 1) // Increment cache hit counter
+		// Remove sector from cache to simulate cache miss
+		vm.cache.Remove(root)
```
n8maninger, line 867: We don't want to remove the sector from the cache here, especially not for a test. A better way would be to add `VolumeManager.ResizeCache(int)` and set the cache size to 0 in the broken tests. There should only be two broken ones: `TestStoreSector` and `TestCheckIntegrity`. 

- In `TestCheckIntegrity`, you have direct access to the storage manager. 
- In `TestStoreSector`, you can get it through `host.Storage()`

host/storage/storage.go
```diff
@@ -956,6 +984,8 @@ func (vm *VolumeManager) PruneSectors() (int, error) {
 
 // NewVolumeManager creates a new VolumeManager.
 func NewVolumeManager(vs VolumeStore, a Alerts, cm ChainManager, log *zap.Logger) (*VolumeManager, error) {
+	// Initialize cache with LRU eviction and a max capacity of 64
+	cache, _ := lru.New[types.Hash256, *[rhpv2.SectorSize]byte](64)
```
n8maninger, line 988: Need to check the error here 
```go
if err != nil {
    return nil, fmt.Errorf("failed to initialize cache: %w", err)
}
```

n8maninger (changes requested):

host/storage/storage.go
```diff
@@ -954,8 +976,19 @@ func (vm *VolumeManager) PruneSectors() (int, error) {
 	return vm.vs.PruneSectors()
 }
 
+// ResizeCache resizes the cache to the given size.
+func (vm *VolumeManager) ResizeCache(size int) error {
```
n8maninger, line 980: no need to return anything

host/storage/storage.go
```diff
@@ -185,7 +192,7 @@ func (vm *VolumeManager) loadVolumes() error {
 			vm.a.Register(alerts.Alert{
 				ID:       frand.Entropy256(),
 				Severity: alerts.SeverityError,
-				Message:  "Failed to open volume",
+				Message:  fmt.Sprintf("Failed to open volume %q", vol.LocalPath),
```
n8maninger, line 195: this line shouldn't be changed

host/storage/storage.go
```diff
@@ -626,7 +633,6 @@ func (vm *VolumeManager) AddVolume(localPath string, maxSectors uint64, result c
 			log.Error("failed to initialize volume", zap.Error(err))
 			alert.Message = "Failed to initialize volume"
 			alert.Severity = alerts.SeverityError
-			alert.Data["error"] = err.Error()
```
n8maninger, line 629: ditto

n8maninger (changes requested):

n8maninger:
Looks like `TestRemoveCorrupt` also needs to have the cache set to `0`

SAHU-01:
> Looks like `TestRemoveCorrupt` also needs to set the cache to `0`. Try running `go test -race -tags=testing ./...` to see if there are any more.
After running `go test -race -tags=testing ./...`  no other packages appear to need to resize the cache to 0. The storage package is properly handling that in its tests.

n8maninger:
Great job on this!

PR #183, opened 2023-10-09
ugprade `renterd`

peterjan:
This PR upgrades the `renterd` dependency and moves over some session code we removed on `renterd`'s side. This PR is necessary to prepare `hostd` for upgrading its `core` dependency.


rhp/v2/rpc_test.go
```diff
@@ -355,7 +355,7 @@ func TestRenew(t *testing.T) {
 		}
 		// upload the sector
 		remainingDuration = contractExpiration - currentHeight
-		price, collateral := rhp2.RPCAppendCost(session.Settings(), remainingDuration)
+		price, collateral := rhp2.RPCAppendCost(*session.Settings(), remainingDuration)
```
peterjan, line 358: the pointers fix themselves here when upgrading the `core` package which uses pointer receivers

peterjan:
closing in favour of https://github.com/SiaFoundation/hostd/pull/185 which includes both

peterjan:
Going to reopen this to make the diff smaller in #185 ... seems like after upgrading `core` the txn signatures become invalid.

n8maninger (approved):

Issue #107, opened 2023-07-11
"Simple" mode for pricing


n8maninger:
### Description

Add a toggle for simple/advanced mode for host configuration. Simple should be the default.

## "Simple" Mode
Visible settings:
- accepting contracts
- address
- storage price
- egress price
- ingress price
- collateral multiplier
- DNS settings
- bandwidth settings

Calculated settings:
- max collateral
  - max collateral would be automatically calculated with the following formula: `12960 * storage price per byte/block * 1 TB * collateral multiplier` 
  - e.g. a storage price of `400 SC/TB/mo` (`92592592593 H/byte/block`)  and collateral multiplier of `2.0` would have a max collateral of `2400 SC`

Hidden settings:
- max duration
- contract price
- base RPC price
- sector access price
- price table validity
- registry settings
- account settings

## Advanced Mode
Would work as it does today. All fields are visible, no calculated fields

alexfreska:
@n8maninger when toggling from advanced to simple would the hidden fields remain the same or get set to some default values, what should those defaults be?

n8maninger:
> @n8maninger when toggling from advanced to simple would the hidden fields remain the same or get set to some default values, what should those defaults be?

I’m leaning towards leaving them untouched. What do you think?

alexfreska:
Yeah I think thats the way to go, lets try that. If it ends up being confusing we could add some UX around edge cases, such as when someone switches to advanced mode changes advanced settings and then toggles back to basic view I could diff the hidden options with a set of defaults and add a warning.

PR #193, opened 2023-10-20
ui: v0.24.0

alexfreska:
- The configuration now has an advanced mode that allows the user to view and change all settings.
- New users will now see an onboarding wizard that prompts the user to complete the necessary setup steps - it also shows the status and progress of each.
- The configuration page now shows the changed status on fields if the user has made a change but the server values were since updated.
- The command palette now includes navigation to configuration sections.
- Fixed an issue where fiat input fields values were not displaying properly.
- Data tables now refresh themselves without user interaction or refocusing the app, even more frequently if a long running operation is in progress.
- The configuration is now much simpler by default, only requiring the user to set essential settings.


n8maninger:
Closes #107

PR #213, opened 2023-11-19
Add API endpoints for Prometheus metrics collection

bustedware:
Added various `GET` endpoints to return metrics formatted for Prometheus


n8maninger:
I'm not crazy about adding specific prometheus endpoints. It's very niche to have as native endpoints. Running a prometheus exporter works fine.

n8maninger:
Let me think on it. If I do end up changing my mind, the endpoints should be grouped under `/integrations/prometheus/*`

bustedware:
If its helpful, neo4j follows a pattern where you enable a prometheus endpoint and it runs on a whole different port than the main service. so perhaps a 3rd option to consider too. thanks

PR #230, opened 2023-12-10
Update core dependency

n8maninger:


n8maninger:
@alexfreska  this will fix the json for chain index

PR #231, opened 2023-12-11
added register alerts

justmert:
Added register alerts, though please check the implementation since i may have missed somethings...


n8maninger:
What is the use case for this endpoint?

n8maninger:
I do not see a need for hostd to support registering external alerts. I will add a `/webhooks/:id/test` endpoint instead

PR #237, opened 2023-12-14
Combine bandwidth metrics

n8maninger:
Reduces the number of simultaneous writers to the SQLite database


n8maninger:
@alexfreska, this will break the bandwidth graph.

The data metrics now look like this:
```json
{
  "data": {
    "ingress": 0,
    "egress": 0
  }
}
```

PR #273, opened 2024-01-14
Add prometheus response format to endpoints

n8maninger:
Adds an internal Prometheus encoder for encoding responses in the Prometheus sample format.

Also adds a `response` path param to some endpoints to support different output formats (i.e. `http://localhost:9980/api/settings?response=prometheus`)

Supported formats:
- `prometheus`
- `json`

Supported endpoints:
- `GET /state/host`
- `GET /state/consensus`
- `GET /settings`
- `GET /metrics`
- `GET /wallet`

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -46,8 +46,8 @@ type (
 		UpdateDDNS(force bool) error
 	}
 
-	// Metrics retrieves metrics related to the host
-	Metrics interface {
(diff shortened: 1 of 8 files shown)


n8maninger:
@bustedware I didn't add all of the endpoints, but this is the direction we should be going wrt supporting Prometheus within hostd.

PR #288, opened 2024-02-01
Update on hostd prometheus metrics endpoints

bustedware:
Submitting an update to the prometheus endpoints for hostd.

### Prometheus Encoder
1. the current encoder for hostd is not compatible with the hostd dashboard in [siagrafana](https://github.com/bustedware/siagrafana) repo
2. the current encoder for hostd shouldn't include timestamps. prometheus is a tool for tracking and measuring various metrics, it does this by periodically checking the system's status, and the specific timing of these checks is usually not critical. it's important to note that prometheus is primarily designed for monitoring purposes, rather than serving as a general purpose time series database.
3. this rolls back the url param and reintroduces a startup param `--prometheus` to enable `/prometheus` endpoints which is what the siagrafana scripts are currently generating prometheus configs with.

### Notes
1. the current hostd encoder should be leveraged in a future refactor of the hostd fork.
2. regarding the url params. i did leave comments on this approach very early in the project and never received feedback https://github.com/SiaFoundation/hostd/pull/213 . so i can include in the refactor switching to something like url param if that is still desired.
3. submitting the PR more as a formality for closing the grant than wanting to merge before refactor


n8maninger:
As I said in your first PR, we would prefer to reduce the configuration and maintenance overhead of Prometheus support in the apps by reusing existing endpoints rather than introducing a separate listener, duplicating endpoints, or adding more configuration variables. We're against adding an additional listener (and port forward) to the apps without good reason.

From our perspective, the ideal outcome would be updating the query parameter implementation we currently have in master with the data and labels you need and then migrating your setup scripts (https://github.com/SiaFoundation/hostd/pull/273#issuecomment-1891053560). This should be simple with the Prometheus encoder we wrote (the timestamp field is optional and won't be included in the data unless explicitly set). The current implementation chose its labels to reduce query complexity, but there are obvious improvements that could be made.

PR #297, opened 2024-02-06
Fix offsets in rpcSectorRoots

n8maninger:


host/contracts/manager.go
```diff
@@ -102,11 +102,7 @@ type (
 	}
 )
 
-func (cm *ContractManager) getSectorRoots(id types.FileContractID, limit, offset int) ([]types.Hash256, error) {
```
ChrisSchinnerl, line 105: Why get rid of the pagination here? Seems like it would slow things down when clients decide to fetch a lot of small batches.

ChrisSchinnerl (approved):

host/contracts/manager.go
n8maninger: The host needs all sector roots to build the proof regardless of the renter's requested range. This enforces that rather than relying on us to always remember it.

PR #308, opened 2024-02-16
Refactor sector prune

n8maninger:
Simplifies the sector pruning code improving performance and readability


n8maninger:
Closes #309

PR #312, opened 2024-02-21
Prometheus: remove id label on hostd_consensus_state_index

bustedware:
The id label on hostd_consensus_state_index is breaking continuity with block height panels and was introduced during the refactor. Removing the label will restore continuity. 

Screenshot 2024-02-21 173716

### DIFF
--- a/api/prometheus.go
+++ b/api/prometheus.go
@@ -56,9 +56,8 @@ func (cs ConsensusState) PrometheusMetric() []prometheus.Metric {
 			}(),
 		},
 		{
-			Name:   "hostd_consensus_state_index",
-			Labels: map[string]any{"id": cs.ChainIndex.ID},
-			Value:  float64(cs.ChainIndex.Height),
+			Name:  "hostd_consensus_state_index",
+			Value: float64(cs.ChainIndex.Height),
 		},
 	}
 }


n8maninger:
`id` is an important piece of consensus information. You can adjust the query to fix it rather than removing the label.

Either
```
min (hostd_consensus_state_index{})
```
Screenshot 2024-02-21 at 2 54 57 PM


If you want to separate multiple instances:
```
min by (name) (hostd_consensus_state_index{})
```
Screenshot 2024-02-21 at 2 54 34 PM

bustedware:
@n8maninger if you're just going to remove it, why have it collected? personally, i don't think, I would ever look at that value. not dismissing its important though.

n8maninger:
> @n8maninger if you're just going to remove it, why have it collected? personally, i don't think, I would ever look at that value. not dismissing its important though.

We use it on our dashboard

bustedware:
@n8maninger can I ask how do you use it in your dashboard? it will help if I understand so I can add to siagrafana. thank you

bustedware:
@n8maninger something else to consider:

https://last9.io/blog/how-to-manage-high-cardinality-metrics-in-prometheus/
> High-label cardinality means that a metric has many unique label combinations. This can have implications for the performance and scalability of Prometheus, as each distinct label combination creates a separate time series in the system. A high cardinality can increase memory usage, CPU usage, longer query times, and higher storage requirements

PR #316, opened 2024-03-01
Add Support for Price Pinning

n8maninger:
- Adds two new endpoints, `[GET] /settings/pinned` and `[PUT] /settings/pinned`
- Adds integrated support for explorer data. If explorer data is disabled in the config, pinning is also disabled.


host/settings/pin/pin.go
```diff
@@ -0,0 +1,262 @@
...
+	defer m.mu.Unlock()
+
+	maxRates := int(m.rateWindow / m.frequency)
+	m.rates = append(m.rates, current)
+	if len(m.rates) >= maxRates {
+		m.rates = m.rates[1:]
+	}
+
+	// skip updating prices if the pinned settings are zero
+	if m.pinned.Storage <= 0 && m.pinned.Ingress <= 0 && m.pinned.Egress <= 0 && m.pinned.MaxCollateral <= 0 {
```
ChrisSchinnerl, line 129: Might be a bit more robust to be explicit about it and use a boolean to enable/disable pinning.
I feel like it might make sense to be able to pin e.g. storage to $0 when pinning is enabled versus considering pinning for storage to be disabled when it is set to 0 and having to update the regular price settings.

host/settings/pin/pin_test.go
```diff
@@ -0,0 +1,269 @@
...
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	if !strings.EqualFold(currency, e.currency) {
+		return 0, errors.New("currency not found")
+	}
+	return e.value, nil
+}
+
+func convertToCurrency(target decimal.Decimal, rate decimal.Decimal) types.Currency {
```
ChrisSchinnerl, line 44: I could probably sleep better if there was a table test for this.

host/settings/pin/pin_test.go
```diff
@@ -0,0 +1,269 @@
...
+		pin.WithLogger(log.Named("pin")))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	initialSettings := sm.Settings()
+	pin := pin.PinnedSettings{
+		Currency: "usd",
+
+		Threshold:     0.1,
```
ChrisSchinnerl, line 124: `10%` is quite a lot. I'd probably do `2-3%` to also trigger in more stable market conditions. Otherwise you might miss out of a good amount of money as a host.

host/settings/pin/pin.go
n8maninger: There's no point to pin to 0. 0 is 0 no matter what currency you're in

host/settings/pin/pin_test.go
n8maninger: The user can set it to whatever they want. I don't think it really matters in the test

persist/sqlite/settings.go
```diff
@@ -10,10 +10,34 @@ import (
...
+	return
+}
+
+// UpdatePinnedSettings updates the host's pinned settings.
+func (s *Store) UpdatePinnedSettings(p pin.PinnedSettings) error {
+	const query = `INSERT INTO host_pinned_settings (id, currency, threshold, storage_price, ingress_price, egress_price, max_collateral) VALUES (0, $1, $2, $3, $4, $5, $6) 
+ON CONFLICT (id) DO UPDATE SET currency=EXCLUDED.currency, threshold=EXCLUDED.threshold, storage_price=EXCLUDED.storage_price, ingress_price=EXCLUDED.ingress_price, egress_price=EXCLUDED.egress_price, max_collateral=EXCLUDED.max_collateral
+RETURNING id;`
+
+	var dummyID int64
```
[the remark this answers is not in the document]
n8maninger: ~~bad~~ habit

ChrisSchinnerl (approved):

api/options.go
```diff
@@ -0,0 +1,104 @@
...
+
+// ServerWithAlerts sets the alerts manager for the API server.
+func ServerWithAlerts(al Alerts) ServerOption {
+	return func(a *api) {
+		a.alerts = al
+	}
+}
+
+// ServerWithWebHooks sets the webhooks manager for the API server.
+func ServerWithWebHooks(w WebHooks) ServerOption {
```
peterjan, line 16: `nit` we treat webhook like filename essentially and write it as `Webhook` (which I think is more correct)

cmd/hostd/main.go
```diff
@@ -350,16 +357,36 @@ func main() {
...
+		api.ServerWithAccountManager(node.accounts),
+		api.ServerWithVolumeManager(node.storage),
+		api.ServerWithRHPSessionReporter(node.sessions),
+		api.ServerWithMetricManager(node.metrics),
+		api.ServerWithSettings(node.settings),
+		api.ServerWithWallet(node.w),
+		api.ServerWithLogger(log.Named("api")),
+	}
+	if node.pinned != nil {
+		// pinner should be nil if explorer data is disabled
```
peterjan, line 382: I don't understand this comment? There doesn't seem to be any checks here that check explorer data being disabled yes or no? I think the `if` is not really necessary here, if `node.pinned` is `nil` we can safely set it on the `api` object.

host/settings/announce_test.go
```diff
@@ -42,7 +42,13 @@ func TestAutoAnnounce(t *testing.T) {
...
 
 	am := alerts.NewManager(webhookReporter, log.Named("alerts"))
-	manager, err := settings.NewConfigManager(dir, hostKey, "localhost:9882", db, node.ChainManager(), node.TPool(), node, am, log.Named("settings"))
+	manager, err := settings.NewConfigManager(settings.WithHostKey(hostKey),
+		settings.WithStore(db),
+		settings.WithChainManager(node.ChainManager()),
+		settings.WithTransactionPool(node.TPool()),
+		settings.WithWallet(node),
+		settings.WithAlertManager(am),
+		settings.WithLog(log.Named("settings")))
```
peterjan, line 51: `nit` `settings` vs `config`

host/settings/pin/pin.go
```diff
@@ -0,0 +1,296 @@
...
+		explorer ExchangeRateRetriever
+		sm       SettingsManager
+
+		frequency  time.Duration
+		rateWindow time.Duration
+
+		mu       sync.Mutex
+		rates    []decimal.Decimal
+		lastRate decimal.Decimal
+		settings PinnedSettings // in-memory cache of pinned settings
```
peterjan, line 79: Oh ok, so I guess because there's an in-memory cache the getter doesn't need a context. I'd still define it on the interface for potential implementations that don't have an in-memory cache 🤷‍♂️  but I see why you didn't add one now.

host/settings/pin/pin_test.go
```diff
@@ -0,0 +1,322 @@
...
+			return fmt.Errorf("failed to convert storage price: %v", err)
+		} else if !storagePrice.Div64(4320).Div64(1e12).Equals(settings.StoragePrice) {
+			return fmt.Errorf("expected storage price %d, got %d", storagePrice, settings.StoragePrice)
+		}
+	}
+
+	if pinned.Ingress.IsPinned() {
+		ingressPrice, err := pin.ConvertCurrencyToSC(decimal.NewFromFloat(pinned.Ingress.Value), rate)
+		if err != nil {
+			return fmt.Errorf("failed to convert storage price: %v", err)
```
peterjan, line 58:
```suggestion
			return fmt.Errorf("failed to convert ingress price: %v", err)
```

host/settings/pin/pin_test.go
```diff
@@ -0,0 +1,322 @@
...
+		} else if !ingressPrice.Div64(1e12).Equals(settings.IngressPrice) {
+			return fmt.Errorf("expected ingress price %d, got %d", ingressPrice, settings.IngressPrice)
+		}
+	}
+
+	if pinned.Egress.IsPinned() {
+		egressPrice, err := pin.ConvertCurrencyToSC(decimal.NewFromFloat(pinned.Egress.Value), rate)
+		if err != nil {
+			return fmt.Errorf("failed to convert storage price: %v", err)
```
peterjan, line 67:
```suggestion
			return fmt.Errorf("failed to convert egress price: %v", err)
```

host/settings/pin/pin_test.go
```diff
@@ -0,0 +1,322 @@
...
+		} else if !egressPrice.Div64(1e12).Equals(settings.EgressPrice) {
+			return fmt.Errorf("expected egress price %d, got %d", egressPrice, settings.EgressPrice)
+		}
+	}
+
+	if pinned.MaxCollateral.IsPinned() {
+		maxCollateral, err := pin.ConvertCurrencyToSC(decimal.NewFromFloat(pinned.MaxCollateral.Value), rate)
+		if err != nil {
+			return fmt.Errorf("failed to convert storage price: %v", err)
```
peterjan, line 76:
```suggestion
			return fmt.Errorf("failed to convert collateral price: %v", err)
```

persist/sqlite/init.sql
```diff
@@ -196,6 +196,20 @@ CREATE TABLE host_settings (
 	sector_cache_size INTEGER NOT NULL DEFAULT 0
 );
 
+CREATE TABLE host_pinned_settings (
+	id INTEGER PRIMARY KEY NOT NULL DEFAULT 0 CHECK (id = 0), -- enforce a single row
```
peterjan, line 200: awesome constraint, going to try and steal this for our current tip

peterjan (commented):
LGTM overall 👍

api/endpoints.go
```diff
@@ -199,6 +200,28 @@ func (a *api) handlePATCHSettings(c jape.Context) {
 	c.Encode(a.settings.Settings())
 }
 
+func (a *api) handleGETPinnedSettings(c jape.Context) {
+	if a.pinned == nil {
+		c.Error(errors.New("pinned settings disabled"), http.StatusNotFound)
```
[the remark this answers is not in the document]
n8maninger: I think 404 is more appropriate since it's disabled

api/options.go
n8maninger: It's `WebHooks` everywhere in `hostd`, we can handle it in a fu.

ChrisSchinnerl (approved):

PR #365, opened 2024-04-16
Add OpenBSD support to diskutils

n8maninger:


internal/disk/usage_openbsd.go
```diff
@@ -0,0 +1,20 @@
...
+import "golang.org/x/sys/unix"
+
+// Usage returns the free and total bytes on the filesystem containing the
+// specified path.
+func Usage(p string) (free, total uint64, err error) {
+	var stat unix.Statfs_t
+	if err := unix.Statfs(p, &stat); err != nil {
+		return 0, 0, err
+	}
+	return stat.F_bfree * uint64(stat.F_bsize), stat.F_blocks * uint64(stat.F_bsize), nil
```
ChrisSchinnerl, line 14: Can we have a small comment on why any of this is necessary and why we can use `F_bsize` and `F_bfree` instead of `Bsize` and `Blocks`? Had to go to the Linux man page to figure out what those mean and am still not sure how that relates to OpenBSD :sweat_smile:

internal/disk/usage_openbsd.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// Drives returns the paths of all drives on Windows. It is a no-op on other systems
```
peterjan, line 17: copy paste (windows)

ChrisSchinnerl (approved):

PR #369, opened 2024-04-22
Remove volume fill

n8maninger:


PiotrZSL:
Ok, so with this change, we got "zero" initialization and random sector selection based on writes.
It solves issue for some time, but thing is that on COW filesystem we need first one (and reusing sectors is welcome, as filesystem will spread a load on hdd), on non-cow filesystem we need second one mainly.

As hostd uses sqlite, and now we got number or writes in database (=info what sectors were actually used), then probably some external application could now lock sector in db, clean it, reset write index, unlock sector.

PiotrZSL:
I just teted this, looks like "ORDER BY vs.sector_writes ASC" is too slow, it takes 0.3s, when normally entire query took 0.001 second. In fact using any sort (like sort by volume id is causing slowdown).

PR #387, opened 2024-05-06
Merge chain manager and tpool in SingleAddressWallet

n8maninger:
@peterjan, is this enough to fix the race condition, or do I need to change the deferred release to only be called when the transaction should be dropped?


peterjan:
@n8maninger unfortunately merging the `tpool` into the manager doesn't fix the race condition. I wasn't sure if that would made you revert the change entirely but I took the liberty to add a commit that release only on error. Feel free to revert that though.

peterjan (approved):
I think maybe adding the tpool to the wallet instead of the chain manager makes for a smaller diff? We have the tpool on the test node already and it would avoid all cases where you have to instantiate a tpool only because we need a chain manager, e.g. in the storage tests.

PR #388, opened 2024-05-07
Bump upload-artifact to v4

n8maninger:


ChrisSchinnerl:
@n8maninger did you verify this works without any additional changes? I think I vaguely remember an issue with this due to artefacts not being allowed to have the same path anymore or something like that.

n8maninger:
Yeah, it broke everything.

PR #408, opened 2024-06-05
Increase timeout when writing merkle proof

ChrisSchinnerl:
Arequipa is receiving large merkle proof responses from then hosts. Some 40MB+ in size. A slow host can't transmit that data within 60 seconds so I increased it to 5 minutes.


n8maninger:
Is `renterd` using v2 for uploads?

ChrisSchinnerl:
@n8maninger No, just for pruning.

PR #430, opened 2024-07-11
Prevent price pin failure from crashing node

n8maninger:
Removes the fatal startup check for explorer configuration and adds an alert instead.


n8maninger:
Closes #430

PR #444, opened 2024-08-13
Update AccountStore interface

peterjan:
While I was browsing the accounts code I found that both `CreditAccountWithContract` and `DebitAccount` don't actually return the account balance because of scoping. Since those return variables weren't being used I figured we might as well get rid of them.


n8maninger (approved):

n8maninger:
Nice find, thanks!

PR #447, opened 2024-08-19
Expose `storage.ErrSectorNotFound` error

peterjan:
Hosts return `ErrHostInternalError` when trying to read a sector that can't be found, we can safely expose that error.

### DIFF
--- a/rhp/v3/rpc_test.go
+++ b/rhp/v3/rpc_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl (changes requested):
I'm pretty sure this has worked before considering the lost sectors we saw in our renter logs. Can we add a test to prevent another regression?

rhp/v3/rpc_test.go
```diff
@@ -150,6 +153,17 @@ func TestAppendSector(t *testing.T) {
 			t.Fatal("downloaded sector doesn't match")
 		}
 	}
+
+	// assert ReadSector exposes ErrSectorNotFound
```
ChrisSchinnerl, line 157: Is there no `ReadSector` test we can extend rather than `TestAppendSector`?

peterjan: Didn't want to add one just for this small regression test 🤷‍♂️ . I don't mind extracting it into a separate test though, Nate be the judge?

PR #463, opened 2024-09-11
Add SQLite3 backup

n8maninger:
Adds an endpoint to backup a running `hostd` node to the local system. Can be used to schedule automatic atomic backups without shutting down the node or locking the database for a significant amount of time.

`[POST] /api/system/sqlite3/backup`

```sh
curl -u :"sia is cool" http://localhost:9980/api/system/sqlite3/backup -X POST -d '{ "path": "/backups/hostd.sqlite3.bak" }'
```


api/endpoints.go
```diff
@@ -579,6 +580,18 @@ func (a *api) handlePUTSystemDir(jc jape.Context) {
 	a.checkServerError(jc, "failed to create dir", os.MkdirAll(req.Path, 0775))
 }
 
+func (a *api) handlePOSTSystemSQLite3Backup(jc jape.Context) {
+	if a.sqlite3Store == nil {
+		jc.Error(errors.New("sqlite3 store not available"), http.StatusNotFound)
```
ChrisSchinnerl, line 585: Missing `return`.

persist/sqlite/store.go
```diff
@@ -65,12 +67,23 @@ func (s *Store) transaction(fn func(*txn) error) error {
...
+	return backupDB(ctx, s.db, destPath)
+}
+
 func sqliteFilepath(fp string) string {
 	params := []string{
 		fmt.Sprintf("_busy_timeout=%d", busyTimeout),
 		"_foreign_keys=true",
 		"_journal_mode=WAL",
 		"_secure_delete=false",
+		"_auto_vacuum=INCREMENTAL",
```
ChrisSchinnerl, line 86: I think this is only useful if you also use `incremental_vacuum` instead of the regular `VACUUM` which doesn't seem to be the case. 

> When the value of auto-vacuum is 2 or "incremental" then the additional information needed to do auto-vacuuming is stored in the database file but auto-vacuuming does not occur automatically at each commit as it does with auto_vacuum=full. In incremental mode, the separate [incremental_vacuum](https://sqlite.org/pragma.html#pragma_incremental_vacuum) pragma must be invoked to cause the auto-vacuum to occur.

[Source](https://sqlite.org/pragma.html#pragma_auto_vacuum)

persist/sqlite/store.go
```diff
@@ -145,6 +158,120 @@ func (s *Store) clearLocks() error {
...
+
+	// initialize the destination conn
+	dc, err := sqlConn(ctx, dest)
+	if err != nil {
+		return fmt.Errorf("failed to create destination connection: %w", err)
+	}
+	defer dc.Close()
+
+	// start the backup
+	backup, err := dc.Backup("main", sc, "main")
```
ChrisSchinnerl, line 214: Considering that `dc` and `sc` are already connections, what are `main` and `main` used for?

ChrisSchinnerl (commented):
Overall this looks good. Only thing I don't love is passing paths to a foreign filesystem through an endpoint but I guess that can't be avoided if we want to allow performing a backup through the UI.

persist/sqlite/store.go
n8maninger: The plan is to eventually implement periodic incremental vacuums. Unfortunately, this needs to be set before you create the database or after a VACUUM.

persist/sqlite/store.go
n8maninger: main is the default schema name in SQLite. So it’s backing up the main schema.

ChrisSchinnerl (approved):

PR #502, opened 2024-11-18
Refactor RHP rate limiting

n8maninger:
Refactors RHP rate limiting to support the new RHP4 listener. Broken out of #490


rhp/listener.go
```diff
@@ -0,0 +1,116 @@
...
+// Listen returns a new listener with optional rate limiting and monitoring.
+func Listen(network, address string, opts ...Option) (net.Listener, error) {
+	l, err := net.Listen(network, address)
+	if err != nil {
+		return nil, err
+	}
+
+	rhp := &rhpListener{
+		l:            l,
+		readLimiter:  rate.NewLimiter(rate.Inf, 0),
```
peterjan, line 108: `nit` could add an extra constructor that returns a `NoOpLimiter` (or `rate.NoLimit()`)

rhp/v2/rpc_test.go
```diff
@@ -59,12 +59,14 @@ func TestSettings(t *testing.T) {
...
 	defer l.Close()
 
-	sh, err := rhp2.NewSessionHandler(l, hostKey, "localhost:9983", node.Chain, node.Syncer, node.Wallet, node.Contracts, node.Settings, node.Volumes, rhp2.WithLog(log))
+	sh, err := rhp2.NewSessionHandler(l, hostKey, "localhost:9983", node.Chain, node.Syncer, node.Wallet, node.Contracts, node.Settings, node.Volumes, log)
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer sh.Close()
 	go sh.Serve()
+	defer sh.Close()
```
peterjan, line 68: copy pasta

rhp/v2/rhp.go
```diff
@@ -287,7 +269,7 @@ func (sh *SessionHandler) LocalAddr() string {
 }
 
 // NewSessionHandler creates a new RHP2 SessionHandler
-func NewSessionHandler(l net.Listener, hostKey types.PrivateKey, rhp3Addr string, cm ChainManager, s Syncer, wallet Wallet, contracts ContractManager, settings SettingsReporter, storage StorageManager, opts ...SessionHandlerOption) (*SessionHandler, error) {
+func NewSessionHandler(l net.Listener, hostKey types.PrivateKey, rhp3Addr string, cm ChainManager, s Syncer, wallet Wallet, contracts ContractManager, settings SettingsReporter, storage StorageManager, log *zap.Logger) (*SessionHandler, error) {
```
peterjan, line 272: Now that options are gone you could ask for the siamux port directly in the parameters and get rid of the error. The `NewSessionHandler` in rhp3 never returns an error so we could also remove it there. Bit of a nit but worth it in my opinion because you make it very clear why an RHP2 session handler needs the RHP3 address and you simplify the setup a bit.

rhp/listener.go
ChrisSchinnerl: `rate` is an external package (https://pkg.go.dev/golang.org/x/time/rate)

rhp/listener.go
```diff
@@ -0,0 +1,116 @@
...
+		net.Conn
+
+		monitor DataMonitor
+	}
+
+	rhpListener struct {
+		l net.Listener
+
+		readLimiter  *rate.Limiter
+		writeLimiter *rate.Limiter
```
ChrisSchinnerl, line 35: Seems like you are setting these fields but not using them?

rhp/v2/rhp.go
n8maninger: Done in #504

cmd/hostd/run.go
```diff
@@ -126,6 +126,37 @@ func startLocalhostListener(listenAddr string, log *zap.Logger) (l net.Listener,
...
+		return 0, fmt.Errorf("failed to split port: %w", err)
+	}
+	port, err := strconv.ParseUint(portStr, 10, 16)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse port: %w", err)
+	}
+	return uint16(port), nil
+}
+
+func normalizeAddress(addr string) (string, error) {
```
n8maninger, line 146: 🤢

n8maninger: Kind of sad, but a relatively straightforward way to solve the issue with the circular dependency on the RHP3 port, settings manager, and bandwidth limiters.

Open to ideas

peterjan (approved):
LGTM 👍

PR #504, opened 2024-11-18
Move settings and price table into config manager

n8maninger:
Refactors RHP2 and RHP3 settings and price tables to be generated in the config manager instead of in the RHP session handler to be consistent with RHP4. Also happens to remove the annoyance of passing in the RHP3 address into the RHP2 handler. 

Broken out of #490.


peterjan (approved):
LGTM 👍 neat cleanup

ChrisSchinnerl (approved):

PR #508, opened 2024-11-19
Fix missing version variable in commit message

n8maninger:


knope.toml
```diff
@@ -20,6 +22,7 @@ type = "PrepareRelease"
 [[workflows.steps]]
 type = "Command"
 command = "git commit -m \"chore: prepare release $version\""
+variables = { "$version" = "Version" }
```
ChrisSchinnerl, line 25: A few lines below this there is 
```
[workflows.steps.variables]
"$version" = "Version"
```

Is that redundant now?

ChrisSchinnerl (commented):
Please open a PR on renterd where you add the same changes after merging this to keep things consistent.

knope.toml
n8maninger: Seems like it’s necessary on each line

PR #510, opened 2024-11-22
Add support for RHP4 accounts

n8maninger:
Broken out of #490 . Requires #507

### DIFF
--- a/persist/sqlite/accounts.go
+++ b/persist/sqlite/accounts.go
@@ -7,11 +7,127 @@ import (
 	"time"
 
 	rhp3 "go.sia.tech/core/rhp/v3"
+	proto4 "go.sia.tech/core/rhp/v4"
 	"go.sia.tech/core/types"
 	"go.sia.tech/hostd/host/accounts"
 	"go.sia.tech/hostd/host/contracts"
 	"go.uber.org/zap"
 )
+
+// RHP4AccountBalance returns the balance of the account with the given ID.
+func (s *Store) RHP4AccountBalance(account proto4.Account) (balance types.Currency, err error) {
+	err = s.transaction(func(tx *txn) error {
+		return tx.QueryRow(`SELECT balance FROM accounts WHERE account_id=$1`, encode(account)).Scan(decode(&balance))
+	})
+	return
+}
+
+// RHP4DebitAccount debits the account with the given ID.
+func (s *Store) RHP4DebitAccount(account proto4.Account, usage proto4.Usage) error {
+	return s.transaction(func(tx *txn) error {
+		var dbID int64
+		var balance types.Currency
+		err := tx.QueryRow(`SELECT id, balance FROM accounts WHERE account_id=$1`, encode(account)).Scan(&dbID, decode(&balance))
+		if err != nil {
+			return fmt.Errorf("failed to query balance: %w", err)
+		}
+
+		total := usage.RenterCost()
+		balance, underflow := balance.SubWithUnderflow(total)
+		if underflow {
+			return fmt.Errorf("insufficient balance")
+		}
+
+		_, err = tx.Exec(`UPDATE accounts SET balance=$1 WHERE id=$2`, encode(balance), dbID)
+		if err != nil {
+			return fmt.Errorf("failed to update balance: %w", err)
+		} else if err := updateV2ContractFunding(tx, dbID, usage); err != nil {
+			return fmt.Errorf("failed to update contract funding: %w", err)
+		}
+		return nil
+	})
+}
+
+// RHP4CreditAccounts credits the accounts with the given deposits and revises
+// the contract.
+func (s *Store) RHP4CreditAccounts(deposits []proto4.AccountDeposit, contractID types.FileContractID, revision types.V2FileContract) (balances []types.Currency, err error) {
+	err = s.transaction(func(tx *txn) error {
+		getBalanceStmt, err := tx.Prepare(`SELECT balance FROM accounts WHERE account_id=$1`)
+		if err != nil {
+			return fmt.Errorf("failed to prepare get balance statement: %w", err)
+		}
+		defer getBalanceStmt.Close()
+
+		updateBalanceStmt, err := tx.Prepare(`INSERT INTO accounts (account_id, balance, expiration_timestamp) VALUES ($1, $2, $3) ON CONFLICT (account_id) DO UPDATE SET balance=EXCLUDED.balance, expiration_timestamp=EXCLUDED.expiration_timestamp RETURNING id`)
+		if err != nil {
+			return fmt.Errorf("failed to prepare update balance statement: %w", err)
+		}
+		defer updateBalanceStmt.Close()
+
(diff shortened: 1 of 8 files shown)


persist/sqlite/accounts.go
```diff
@@ -7,12 +7,128 @@ import (
...
+		var balance types.Currency
+		err := tx.QueryRow(`SELECT id, balance FROM accounts WHERE account_id=$1`, encode(account)).Scan(&dbID, decode(&balance))
+		if err != nil {
+			return fmt.Errorf("failed to query balance: %w", err)
+		}
+
+		total := usage.RenterCost()
+		balance, underflow := balance.SubWithUnderflow(total)
+		if underflow {
+			return fmt.Errorf("insufficient balance")
```
peterjan, line 38:
```suggestion
			return errors.New("insufficient balance")
```

persist/sqlite/accounts.go
```diff
@@ -7,12 +7,128 @@ import (
...
 	"go.sia.tech/core/types"
 	"go.sia.tech/hostd/host/accounts"
 	"go.sia.tech/hostd/host/contracts"
 	"go.uber.org/zap"
 )
 
+// RHP4AccountBalance returns the balance of the account with the given ID.
+func (s *Store) RHP4AccountBalance(account proto4.Account) (balance types.Currency, err error) {
+	err = s.transaction(func(tx *txn) error {
+		return tx.QueryRow(`SELECT balance FROM accounts WHERE account_id=$1`, encode(account)).Scan(decode(&balance))
```
peterjan, line 20: Do we want to handle `sql.ErrNoRows` and return an `ErrAccountNotFound`?

persist/sqlite/accounts.go
```diff
@@ -7,12 +7,128 @@ import (
...
+			if err != nil && !errors.Is(err, sql.ErrNoRows) {
+				return fmt.Errorf("failed to get balance: %w", err)
+			} else if errors.Is(err, sql.ErrNoRows) {
+				createdAccounts++
+			}
+
+			balance = balance.Add(deposit.Amount)
+
+			var accountDBID int64
+			err = updateBalanceStmt.QueryRow(encode(deposit.Account), encode(balance), encode(time.Now().Add(90*24*time.Hour))).Scan(&accountDBID)
```
peterjan, line 99: `90*24*time.Hour` could be a constant

persist/sqlite/accounts.go
```diff
@@ -7,12 +7,128 @@ import (
...
+			if err != nil {
+				return fmt.Errorf("failed to update balance: %w", err)
```
peterjan, line 101: `nit` the account ID might be useful here

persist/sqlite/contracts.go
```diff
@@ -1121,10 +1118,32 @@ func updateV2ContractUsage(tx *txn, contractDBID int64, usage proto4.Usage) erro
 		} else if err := updateCollateralMetrics(types.ZeroCurrency, usage.RiskedCollateral, false, incrementCurrencyStat); err != nil {
 			return fmt.Errorf("failed to update collateral metrics: %w", err)
 		}
+	} else if status == contracts.V2ContractStatusSuccessful || status == contracts.V2ContractStatusRenewed {
```
peterjan, line 1121: `nit` personally prefer a switch

peterjan (commented):
left some nits but good to go imo 👍

persist/sqlite/accounts.go
```diff
@@ -212,6 +348,65 @@ func contractFunding(tx *txn, accountID int64) (fund []fundAmount, err error) {
 	return
 }
 
+// updateV2ContractFunding distributes account usage to the contracts that funded
+// the account.
+func updateV2ContractFunding(tx *txn, accountID int64, usage proto4.Usage) error {
```
ChrisSchinnerl, line 353: This seems a bit complicated for a solution that results in inaccurate information. It might be simpler and more accurate to always credit the last contract that was used for funding an account.
Or to not associate usage with contracts and instead with accounts.

ChrisSchinnerl: Also the name is not consistent with `updateContractUsage` which seems to be the same thing but for v1?

persist/sqlite/accounts.go
```diff
@@ -265,7 +460,7 @@ func updateContractUsage(tx *txn, accountID int64, usage accounts.Usage, log *za
 			return fmt.Errorf("failed to decrement account funding: %w", err)
 		}
 
-		if contract.Status == contracts.ContractStatusActive || contract.Status == contracts.ContractStatusPending {
+		if contract.Status == contracts.ContractStatusActive {
```
ChrisSchinnerl, line 463: What triggered this change? Isn't this pretty much guaranteed to miss metrics since we start using accounts before the contract appears on chain?

persist/sqlite/accounts_test.go
```diff
@@ -0,0 +1,165 @@
...
+			t.Fatalf("expected storage %v, got %v", earned.Storage, m.Revenue.Earned.Storage)
+		case m.Revenue.Earned.RPC != earned.RPC:
+			t.Fatalf("expected RPC %v, got %v", earned.RPC, m.Revenue.Earned.RPC)
+		}
+	}
+
+	sk := types.GeneratePrivateKey()
+	account := proto4.Account(sk.PublicKey())
+
+	// deposit funds
```
ChrisSchinnerl, line 80: I think `RHP4AccountBalance` is not called at all and we should probably also test the edge cases of withdrawing from an account and fetching the balance of an account before it even exists.

persist/sqlite/accounts.go
n8maninger: account ID doesn't matter if the query fails. It's unlikely to be related to the account ID.

persist/sqlite/accounts.go
n8maninger: Associating usage with an account is unacceptable UX.

persist/sqlite/accounts.go
n8maninger: The contract metrics are added to the global metrics when the status changes from pending -> active, which is consistent with the contract revision PR that just merged. This was a necessary change when finalization was a thing. Now that it's not we could revert it, but it does simplify the edge-cases around pending->reverted->confirmed.

persist/sqlite/accounts.go
ChrisSchinnerl: The UX I imagine is 
1. click on contract
2. contract shows you contract spending and a list of accounts and amounts that it was used to fund with a number indicating the amount. Most of the time that will be a single account.
3. you click the account to get detailed spending metrics for it and a list of contracts that were used to fund it for you to seamlessly navigate between contracts and accounts

If you spread it out across contracts you end up with some awkward UX depending on whether you also include expired contracts or not. Either you increase the spending of contracts which are expired which isn't great or you potentially end up seemingly spending more money from a contract than it contained to begin with.

n8maninger: That UX is significantly more complex than associating it with the contract  in the first place, which is a more accurate representation and what users are expecting.

ChrisSchinnerl (approved):

PR #531, opened 2024-12-15
fix(contracts,sqlite): Fix contract account usage calculation

n8maninger:
`incrementContractUsage` is called from `updateContractUsage` for `distributeRHP3AccountUsage` effectively doubling the contract usage columns.

### DIFF
--- a/persist/sqlite/migrations.go
+++ b/persist/sqlite/migrations.go
@@ -12,6 +12,12 @@ import (
 	"go.uber.org/zap"
 )
 
+// migrateVersion37 recalculates the contract metrics to remove the pending contract
+// values from the metrics.
+func migrateVersion37(tx *txn, log *zap.Logger) error {
+	return recalcContractMetrics(tx, log)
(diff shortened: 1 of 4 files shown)


ChrisSchinnerl (approved):

persist/sqlite/migrations.go
```diff
@@ -12,6 +12,12 @@ import (
 	"go.uber.org/zap"
 )
 
+// migrateVersion37 recalculates the contract metrics to remove the pending contract
+// values from the metrics.
+func migrateVersion37(tx *txn, log *zap.Logger) error {
+	return recalcContractMetrics(tx, log)
```
peterjan, line 18: I noticed you decode registry read/write in your v2 contracts query (see `recalc.go:171`)

n8maninger: 💪🏻

ChrisSchinnerl (approved):

PR #540, opened 2024-12-19
Attempt to upgrade existing configs instead of erroring

n8maninger:
v2.0.0 has a number of config changes that are incompatible with existing configs. Currently, `hostd` will exit if the config contains unexpected fields. This adds support for upgrading an existing v1.1.2 config file to the v2.0.0 config format.

### DIFF
--- a/cmd/hostd/main.go
+++ b/cmd/hostd/main.go
@@ -103,8 +103,7 @@ func tryLoadConfig() {
 	}
 
 	if err := config.LoadFile(configPath, &cfg); err != nil && !errors.Is(err, os.ErrNotExist) {
-		fmt.Println("failed to load config file:", err)
-		os.Exit(1)
+		checkFatalError("failed to load config file", err)
 	}
 }
 
--- a/cmd/hostd/main.go
+++ b/cmd/hostd/main.go
@@ -18,7 +18,6 @@ import (
 	"go.sia.tech/hostd/persist/sqlite"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-	"gopkg.in/yaml.v3"
 	"lukechampine.com/flagg"
 )
 
@@ -65,7 +64,6 @@ var (
 			},
 		},
 		Log: config.Log{
-			Path:  os.Getenv(logFileEnvVar), // deprecated. included for compatibility.
 			Level: "info",
 			File: config.LogFile{
 				Enabled: true,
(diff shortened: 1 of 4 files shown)


cmd/hostd/main.go
```diff
@@ -104,20 +102,8 @@ func tryLoadConfig() {
...
-	f, err := os.Open(configPath)
-	checkFatalError("failed to open config file", err)
-	defer f.Close()
-
-	dec := yaml.NewDecoder(f)
-	dec.KnownFields(true)
-
-	if err := dec.Decode(&cfg); err != nil {
-		fmt.Println("failed to decode config file:", err)
+	if err := config.LoadFile(configPath, &cfg); err != nil && !errors.Is(err, os.ErrNotExist) {
```
peterjan, line 105: `nit` since we don't really use `os.ErrNotExists` we might as well have `config.LoadFile` obfuscate it so the caller doesn't need to worry about it

config/compat.go
```diff
@@ -0,0 +1,172 @@
...
+			// Path is the path of the log file.
+			Path string `yaml:"path,omitempty"`
+		} `yaml:"file,omitempty"`
+	} `yaml:"log,omitempty"`
+}
+
+// updateConfigV112 updates the config file from v1.1.2 to the latest version.
+// It returns an error if the config file cannot be updated.
+func updateConfigV112(fp string, r io.Reader, cfg *Config) error {
+	dec := yaml.NewDecoder(r)
```
peterjan, line 63: are you purposefully not setting `KnownFields` here?

config/compat.go
```diff
@@ -0,0 +1,172 @@
...
+	cfg.HTTP.Password = old.HTTP.Password
+	cfg.Syncer.Address = old.Consensus.GatewayAddress
+	cfg.Syncer.Bootstrap = old.Consensus.Bootstrap
+	cfg.Syncer.Peers = old.Consensus.Peers
+	cfg.Explorer.Disable = old.Explorer.Disable
+	cfg.Explorer.URL = old.Explorer.URL
+	cfg.RHP2.Address = old.RHP2.Address
+	cfg.RHP3.TCPAddress = old.RHP3.TCPAddress
+	cfg.Log.Level = old.Log.Level
+	if cfg.Log.File.Path != "" {
```
peterjan, line 146: should this be `old.Log.File.Path`

config/compat.go
```diff
@@ -0,0 +1,172 @@
...
+			Format  string `yaml:"format,omitempty"`
+			// Path is the path of the log file.
+			Path string `yaml:"path,omitempty"`
+		} `yaml:"file,omitempty"`
+	} `yaml:"log,omitempty"`
+}
+
+// updateConfigV112 updates the config file from v1.1.2 to the latest version.
+// It returns an error if the config file cannot be updated.
+func updateConfigV112(fp string, r io.Reader, cfg *Config) error {
```
peterjan, line 62: this function is not acid and will erase the contents of the existing (old) yaml file if we fail to encode the updated config, not sure if you want to change that but wanted to point it out regardless

cmd/hostd/main.go
```diff
@@ -104,20 +102,8 @@ func tryLoadConfig() {
...
+		fmt.Println("failed to load config file:", err)
```
peterjan, line 106: I remember reviewing a `hostd` PR that logged startup errors to stderr. I think failing to load config might be an inconsistent that's either sneaking back in or was overlooked at the time.

cmd/hostd/main.go
n8maninger: I want to leave it up to the caller what to do instead of turning load into an effective `TryLoad`

config/compat.go
n8maninger: Changed my mind. I think you're right, we should fail if the config is invalid.

PR #546, opened 2024-12-29
Use standard directories

n8maninger:
Uses standard locations for application data instead of the current directory. This brings `hostd` in line with other system services and makes it easier to manage application data. This should fall back to `PWD` if a database file already exists.

#### Linux, FreeBSD, OpenBSD
- Configuration: `/etc/hostd`
- Data directory: `/var/lib/hostd`

#### macOS
- Configuration: `~/Library/Application Support/hostd`
- Data directory: `~/Library/Application Support/hostd`

#### Windows
- Configuration: `%APPDATA%\SiaFoundation\hostd`
- Data directory: `%APPDATA%\SiaFoundation\hostd`

#### Docker
- Configuration: `/data`
- Data directory: `/data`

### DIFF
--- a/.changeset/use_default_data_directories.md
+++ b/.changeset/use_default_data_directories.md
@@ -1,0 +1,19 @@
+---
+default: patch
+---
+
+# Use standard locations for application data
+
+Uses standard locations for application data instead of the current directory. This brings `hostd` in line with other system services and makes it easier to manage application data.
+
+#### Linux, FreeBSD, OpenBSD
+- Configuration: `/etc/hostd`
+- Data directory: `/var/lib/hostd`
+
+#### macOS
+- Configuration: `~/Library/Application Support/hostd`
+- Data directory: `~/Library/Application Support/hostd`
+
+#### Windows
+- Configuration: `%APPDATA%\SiaFoundation\hostd`
+- Data directory: `%APPDATA%\SiaFoundation\hostd`
--- a/.changeset/use_default_data_directories.md
+++ b/.changeset/use_default_data_directories.md
@@ -7,13 +7,17 @@ default: major
 Uses standard locations for application data instead of the current directory. This brings `hostd` in line with other system services and makes it easier to manage application data.
 
 #### Linux, FreeBSD, OpenBSD
-- Configuration: `/etc/hostd`
+- Configuration: `/etc/hostd/hostd.yml`
 - Data directory: `/var/lib/hostd`
 
 #### macOS
-- Configuration: `~/Library/Application Support/hostd`
+- Configuration: `~/Library/Application Support/hostd.yml`
 - Data directory: `~/Library/Application Support/hostd`
 
 #### Windows
-- Configuration: `%APPDATA%\SiaFoundation\hostd`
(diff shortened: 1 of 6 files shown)


cmd/hostd/run.go
```diff
@@ -41,6 +42,32 @@ import (
...
 
+func defaultDataDirectory(fp string) string {
+	// use the provided path if it's not empty
+	if fp != "" {
+		return fp
+	}
+
+	// check for databases in the current directory
+	if _, err := os.Stat("hostd.db"); err == nil {
+		return ""
```
peterjan, line 53: `return "."` to be consistent with :67

cmd/hostd/main.go
```diff
@@ -94,17 +95,46 @@ func openBrowser(url string) error {
...
+	return paths
+}
+
+// tryLoadConfig tries to load the config file. It will try multiple locations
+// based on GOOS starting with PWD/hostd.yml. If the file does not exist, it will
+// try the next location. If an error occurs while loading the file, it will
+// print the error and exit. If the config is successfully loaded, the path to
+// the config file is returned.
+func tryLoadConfig() string {
+	for _, fp := range tryConfigPaths() {
```
peterjan, line 130: Should we change the order of the paths in `tryConfigPaths` and have it return the OS specific paths first? Did you consider either logging what config file gets loaded or print a warning of multiple config files are present?

peterjan (dismissed):
LGTM 👍

cmd/hostd/main.go
n8maninger: I'm not sure about the ordering, but the goal was to be consistent in prioritizing most specific to least specific when deciding which setting to use. The priorities are as follows:

1. CLI Flags
2. Environment Variables
3. Config file
  3.1 PWD
  3.2 Data directory
  3.3 OS path

cmd/hostd/run.go
```diff
@@ -41,6 +42,32 @@ import (
...
+	} else if _, err := os.Stat("hostd.sqlite3"); err == nil {
+		return ""
+	}
+
+	// default to the operating system's application directory
+	switch runtime.GOOS {
```
ChrisSchinnerl, line 59: I'd consider moving this whole switch statement into a helper method to avoid duplicating it 3 times. For the one case where we want to be compatible with the old location we can add a dedicated
```
switch runtime.GOOS {
  case "linux", "freebsd", "openbsd":
    // add old default
}
```
and still save a bunch of lines

.changeset/use_default_data_directories.md
```diff
@@ -0,0 +1,19 @@
+---
+default: patch
```
ChrisSchinnerl, line 2: `major`? Behavior of default `hostd config` command changed.

cmd/hostd/run.go
n8maninger: The three are subtly different:

1. `run.go:59` returns `/var/lib/hostd` on Linux
2. `config.go:258` returns `/etc/hostd/hostd.yml` on Linux
3. `main.go:110`, as you said, returns multiple paths

ChrisSchinnerl (approved):

peterjan (approved):

PR #565, opened 2025-01-09
rhp2, rhp3: remove internal error

n8maninger:


peterjan (approved):
I was going to propose that in the meeting but figured you would never ever go for it 😅

PR #583, opened 2025-01-22
Increase volume concurrency

n8maninger:
Increases the sleep time between batches when performing background volume actions. This improves concurrency and usability of the host during these background tasks at the cost of performance. Migrating data from sectors will take at least 5 hours per TiB. IO from slow disks will cause this to be longer since atomicity is very important in migration. This is more performant than explicitly setting an `EXCLUSIVE` transaction isolation for heavy write transactions.


ChrisSchinnerl (approved):
Ngl I was a bit confused about the test but when I ran it once with the 1ms jitter I understood 😅

peterjan (approved):

PR #598, opened 2025-02-10
Default to SiaScan for explorer

n8mgr:
When migrating the config, it's possible for the explorer URL to be set to the zero value causing an error when the user tries to update their settings. This ensures the URL will properly default to SiaScan when unset.

### DIFF
--- a/cmd/hostd/run.go
+++ b/cmd/hostd/run.go
@@ -449,7 +449,7 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 		api.WithSQLite3Store(store),
 	}
 	if !cfg.Explorer.Disable {
-		if len(cfg.Explorer.URL) == 0 {
+		if cfg.Explorer.URL == "" {
 			cfg.Explorer.URL = "https://api.siascan.com"
 		}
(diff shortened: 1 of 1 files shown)


peterjan (approved):

cmd/hostd/run.go
```diff
@@ -449,6 +449,9 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 		api.WithSQLite3Store(store),
 	}
 	if !cfg.Explorer.Disable {
+		if cfg.Explorer.URL == "" {
+			cfg.Explorer.URL = "https://api.siascan.com"
```
peterjan, line 453: I would've probably defaulted in `LoadFile` and added a comment why the default is necessary. If you say "migration" do you refer to `updateConfigV112`? If so, shouldn't we check for empty explorer URL in that function as well?

n8maninger: Yeah, it seems to be an issue configs migrated with `updateConfigV112`. The field gets explicitly set to "",  which ends up overriding the default when in reality we wanted it unset.

n8maninger: Im going to close this for now and consider restructuring the way we handle default values

PR #602, opened 2025-02-11
Update core deps

ChrisSchinnerl:
Second try. Now with the fixed version.


host/contracts/update.go
```diff
@@ -466,10 +467,14 @@ func buildContractState(tx UpdateStateTx, u stateUpdater, revert bool, log *zap.
 		case rev != nil:
 			if revert {
 				state.Revised = append(state.Revised, fce)
-				log.Debug("revised contract", zap.Uint64("current", rev.FileContract.RevisionNumber), zap.Uint64("revised", fce.FileContract.RevisionNumber))
+				log.Debug("revised contract", zap.Uint64("current", rev.RevisionNumber), zap.Uint64("revised", fce.FileContract.RevisionNumber))
 			} else {
-				state.Revised = append(state.Revised, *rev)
-				log.Debug("revised contract", zap.Uint64("current", fce.FileContract.RevisionNumber), zap.Uint64("revised", rev.FileContract.RevisionNumber))
+				state.Revised = append(state.Revised, types.FileContractElement{
```
ChrisSchinnerl, line 472: @lukechampine is this sane?

@n8maninger I noticed that theoretically we don't need the state element in `applyV2ContractRevision` since it redundantly updates the proof and index. However, a `types.V2FileContract` is still not enough since we don't have the contract's id. 
I played around with two ideas.
1. Updating `StateChanges.Revised` to a `map[types.FileContractID]types.FileContractElement` to get the ID later
2. Introducing the reduced types `RevisedContract` and `RevisedV2Contract` which wrap `FileContract` and `V2FileContract` respectively and add an ID. 

I don't love both approaches since 1. introduces a map that needs to be initialized and 2. adds new types just for this. So maybe this is the cleanest way to achieve what we need after all. Either that or a helper in `core` that does the same thing.

n8maninger: Adding a type and changing the query only update the revision would reduce IO. I'd probably lean towards that even with the unfortunate side effect of adding a new type.

host/contracts/update.go
```diff
@@ -450,8 +450,9 @@ func (cm *Manager) ProcessActions(index types.ChainIndex) error {
 }
 
 // buildContractState helper to build state changes from a state update
-func buildContractState(tx UpdateStateTx, u stateUpdater, revert bool, log *zap.Logger) (state StateChanges) {
-	u.ForEachFileContractElement(func(fce types.FileContractElement, created bool, rev *types.FileContractElement, resolved, valid bool) {
+func buildContractState(tx UpdateStateTx, fces []consensus.FileContractElementDiff, v2Fces []consensus.V2FileContractElementDiff, revert bool, log *zap.Logger) (state StateChanges) {
+	for _, diff := range fces {
```
n8maninger, line 454: Be careful of returns in the diff loops `!relevant` should be `continue` now

host/contracts/update.go
```diff
@@ -450,8 +450,9 @@ func (cm *Manager) ProcessActions(index types.ChainIndex) error {
...
+		fce, rev, created, resolved, valid := diff.FileContractElement, diff.Revision, diff.Created, diff.Resolved, diff.Valid
 		log := log.With(zap.Stringer("contractID", fce.ID))
 		if relevant, err := tx.ContractRelevant(types.FileContractID(fce.ID)); err != nil {
 			log.Fatal("failed to check contract relevance", zap.Error(err))
```
n8maninger, line 458: Could be a follow up, but this can return an error now instad of panicking.

host/contracts/update.go
```diff
@@ -485,11 +490,11 @@ func buildContractState(tx UpdateStateTx, u stateUpdater, revert bool, log *zap.
 		default:
 			log.Fatal("unexpected contract state", zap.Bool("resolved", resolved), zap.Bool("valid", valid), zap.Bool("created", created), zap.Bool("revised", rev != nil), zap.Stringer("contractID", fce.ID))
 		}
-	})
+	}
 
-	u.ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
+	for _, diff := range v2Fces {
```
n8maninger, line 495: Ditto on `return`

n8maninger (changes requested):

n8maninger (approved):

PR #622, opened 2025-03-01
Update README.md

foegra:
in docker compose - it has to be hostd, not host


n8maninger:
It does not have to be hostd -- the service name is up to the user

foegra:
if You leave it "host" - Dockge will see container as "stopped" always, which is wrong

PR #637, opened 2025-03-21
Improve migrate performance

n8maninger:
Improve migrate performance when removing large volumes. The old query took on ~2s to complete, while the split query takes ~3ms. This should help reduce database is locked failures for active hosts.

### DIFF
--- a/persist/sqlite/volumes.go
+++ b/persist/sqlite/volumes.go
@@ -526,12 +526,19 @@ func emptyLocation(tx *txn) (loc storage.SectorLocation, err error) {
 // emptyLocationForMigration returns an empty location in a writable volume. If there is no
 // space available, ErrNotEnoughStorage is returned.
 func emptyLocationForMigration(tx *txn, volumeID int64, maxIndex uint64) (loc storage.SectorLocation, err error) {
+	loc, err = emptyLocation(tx)
+	if !errors.Is(err, storage.ErrNotEnoughStorage) {
+		return // either a db error or a valid location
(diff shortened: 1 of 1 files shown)


peterjan (approved):
No benchmark but I'm sure you tested it manually.

persist/sqlite/volumes.go
```diff
@@ -526,12 +526,19 @@ func emptyLocation(tx *txn) (loc storage.SectorLocation, err error) {
 // emptyLocationForMigration returns an empty location in a writable volume. If there is no
 // space available, ErrNotEnoughStorage is returned.
 func emptyLocationForMigration(tx *txn, volumeID int64, maxIndex uint64) (loc storage.SectorLocation, err error) {
+	loc, err = emptyLocation(tx)
+	if !errors.Is(err, storage.ErrNotEnoughStorage) {
+		return // either a db error or a valid location
+	} else if maxIndex == 0 {
+		err = storage.ErrNotEnoughStorage
+		return // no space in current volume
+	}
```
ChrisSchinnerl, line 535: Did you mean to exit early when `emptyLocation` works? As the code is right now I'm slightly confused because both `emptyLocation` as well as `emptyLocationForMigration` increment `sector_writes` and you don't ever use the result of `emptyLocation` but still don't do `_, err = emptyLocation(tx)` to indicate all you are actually interested in is the `err`.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #655, opened 2025-04-19
Improve renewal RPC performance

n8maninger:
Adds an additional table to map sector roots to a contract. This helps improve the performance of contract renewals. Previously when renewing, every sector root record had to be updated with the new foreign key. This caused the renew RPC to take up to 30s for contracts over 5TiB. With this change only a single record needs to be updated when renewing.

```
BenchmarkRenewContract/1000_sectors-16              2830            466750 ns/op       1000 sectors         32214 B/op        411 allocs/op
BenchmarkRenewContract/10000_sectors-16             2902            475401 ns/op      10000 sectors         32181 B/op        411 allocs/op
BenchmarkRenewContract/100000_sectors-16            3141            478208 ns/op     100000 sectors         32149 B/op        411 allocs/op
BenchmarkRenewContract/1000000_sectors-16           2794            493699 ns/op         1000000 sectors          32187 B/op        411 allocs/op
```

This is a significant change to sector management and should be thoroughly tested before release.

### DIFF
--- a/.changeset/improve_performance_of_renewing_large_contracts.md
+++ b/.changeset/improve_performance_of_renewing_large_contracts.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Improved performance when renewing large contracts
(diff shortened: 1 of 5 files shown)


n8maninger:
Replaced by #807

PR #662, opened 2025-04-28
update deps

n8maninger:
Updates to the latest coreutils and updates tests and call sites that expect the syncer to work to use the mock syncer since BroadcastTransaction will now error when there are no peers connected.


peterjan (approved):

host/contracts/update.go
```diff
@@ -212,9 +212,11 @@ func (cm *Manager) ProcessActions(index types.ChainIndex) error {
 		formationSet, err := tryFormationBroadcast(cm.chain, formationSet)
 		if err != nil {
 			log.Error("failed to add formation transaction to pool", zap.Error(err))
+		} else if err := cm.syncer.BroadcastTransactionSet(formationSet); err != nil {
+			log.Error("failed to broadcast transaction set", zap.Error(err))
+		} else {
+			log.Debug("rebroadcast formation transaction", zap.String("transactionID", formationTxn.ID().String()))
```
ChrisSchinnerl, line 218: (nit) for future reference. You can use `zap.Stringer`

ChrisSchinnerl (approved):

PR #664, opened 2025-04-29
sqlite: remove jitter sleep and transaction retries

n8maninger:
Now that the database is limited to a single conn there is no need to retry transactions.


peterjan (approved):
In SQLite, wouldn't you still need to retry a transaction (sometimes) if you use the same connection in parallel?

n8maninger:
> In SQLite, wouldn't you still need to retry a transaction (sometimes) if you use the same connection in parallel?

Calling `Begin` either creates a new conn or reuses a free conn. Since we call `SetMaxOpenConns` sql.DB's internal mutex _should_ prevent multiple conns existing at a time. Individual conns are not thread safe, so we shouldn't ever be using them in parallel.

PR #674, opened 2025-05-18
Add connectivity checks

n8maninger:
Periodically tests the hosts connection using the SiaScan troubleshooting API. This provides alerts to hosts if for some reason their node is not connectable. This system relies on a central server for the test and may return occasional false positives. The server can be overridden by setting `Config.Explorer.URL`


explorer/troubleshoot.go
```diff
@@ -0,0 +1,107 @@
...
+		Errors   []string `json:"errors"`
+		Warnings []string `json:"warnings"`
+	}
+
+	// RHP4Result is the result of testing a host's RHP4 endpoint. It contains
+	// the results of the connection, handshake, and scan, as well as any errors
+	// or warnings that occurred during the test.
+	RHP4Result struct {
+		NetAddress        chain.NetAddress `json:"netAddress"`
+		ResolvedAddresses []string         `json:"resolvedAddress"`
```
peterjan, line 71:
```suggestion
		ResolvedAddresses []string         `json:"resolvedAddresses"`
```

peterjan (dismissed):

PR #696, opened 2025-05-28
Account for window size in handleRPCRenew when checking for ErrAfterV2Hardfork

ChrisSchinnerl:


rhp/v3/rpc.go
```diff
@@ -293,7 +293,7 @@ func (sh *SessionHandler) handleRPCRenew(s *rhp3.Stream, log *zap.Logger) (contr
 	renewal := renewalTxn.FileContracts[0]
 
 	// prevent forming v1 contracts with proof windows after the v2 hardfork
-	if renewal.WindowStart >= cs.Network.HardforkV2.RequireHeight {
+	if renewal.WindowStart >= cs.Network.HardforkV2.RequireHeight-pt.WindowSize {
```
n8maninger, line 296: Switch this to WindowEnd instead of subbing WindowSize

PR #729, opened 2025-06-12
Always enable quic

n8maninger:
Adds a SiaMux or QUIC listener if one is not specified.

### DIFF
--- a/cmd/hostd/run.go
+++ b/cmd/hostd/run.go
@@ -432,7 +432,17 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	defer vm.Close()
 
 	var rhp4PortStr string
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (dismissed):

cmd/hostd/run.go
```diff
@@ -432,7 +432,18 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
 	defer vm.Close()
 
 	var rhp4PortStr string
+	var hasQuic bool
+	var hasSiamux bool
```
peterjan, line 436: `hasSiamux` is always false

peterjan (changes requested):
since Chris already approved it, I'm requesting changes to ensure we don't accidentally merge it

Issue #742, opened 2025-06-19
Failed to test connection


migoldfinger:
### Current Behavior

2025-06-19T23:29:25+02:00       ERROR   settings        failed to test connection       {"error": "failed to test connection: failed to send request: Post \"https://api.siascan.com/troubleshoot\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"}

### Expected Behavior

Test succeeded

### Steps to Reproduce

1) Start hostd

### Version

v2.3.2

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

windows

### Anything else?

Everything else seems to work fine.

kocoten1992:
related https://github.com/SiaFoundation/hostd/issues/728

PR #749, opened 2025-06-25
Don't print result of `TestConnection` error on shutdown

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/hostd/issues/742

### DIFF
--- a/explorer/connectivity/connectivity.go
+++ b/explorer/connectivity/connectivity.go
@@ -2,6 +2,7 @@ package connectivity
 
 import (
 	"context"
+	"errors"
 	"math"
 	"time"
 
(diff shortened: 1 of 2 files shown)


explorer/connectivity/connectivity.go
```diff
@@ -131,7 +132,9 @@ func NewManager(hostKey types.PublicKey, settings Settings, explorer Explorer, o
 			select {
 			case <-time.After(nextTestTime):
 				result, ok, err := m.TestConnection(context.Background())
```
n8maninger, line 134: This should be using the threadgroup context from above

n8maninger (approved):

Issue #772, opened 2025-07-01
Add alerts for certificate expiration


n8maninger:
> Would it be appropriate to suggest here that hostd should report warning alerts at 30 days till expiry and error alerts a few days before?  

 _Originally posted by @timrettop in [#767](https://github.com/SiaFoundation/hostd/issues/767#issuecomment-3024737969)_

PR #776, opened 2025-07-04
Adding expiration Logging and Alerting to Local Certificate Provider

timrettop:
Here is my first attempt at #772.  I appreciate any feedback.

This uses a constant value for the expiration period alert, starting at 29 days until expiry.   I'm not sure that it needs to be configurable.  

I tested this by building with that value set to 88 days as I have a valid cert thats expiring in 87 days and a new cert (expiring in 89 days). 
Testing Results: 
The alert is registered shortly after starting.  
Dismissing the alert, the alert correctly returns within a couple minutes. 
Replacing the certificate and leaving the alert displaying, after the certificate is loaded from disk (check debug logs for cert expiration date) the alert is correctly dismissed.  

I initially used the Data field in the alerts to provide the expiration date of the certificate, however that was not working, maybe only certain alerts are configured to display data in the UI?

### DIFF
--- a/certificates/providers/local/local.go
+++ b/certificates/providers/local/local.go
@@ -51,7 +51,7 @@ func (p *localProvider) GetCertificate(ctx context.Context) (*tls.Certificate, e
 
 		p.cert = &cert
 		p.lastLoad = time.Now()
-		p.CheckCertificateExpiry(p.cert, p.alerter, p.log)
+		p.checkCertificateExpiry()
 	}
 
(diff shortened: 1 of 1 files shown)


timrettop:
@n8maninger Thank you! I've applied your feedback and added handling if the expiration lapses as well.  Ready for review. 
I also looked into the web project to see how to properly display the alert data and will submit a separate pull request for that.

n8maninger (approved):

PR #791, opened 2025-07-30
Update core and coreutils

n8maninger:


internal/integration/rhp/v4/rhp4_test.go
```diff
@@ -204,7 +204,284 @@ func TestFormContract(t *testing.T) {
...
+		}
+	})
+
+	t.Run("valid refresh - no renter funding", func(t *testing.T) {
+		expirationHeight := cm.Tip().Height + 50 + proto4.ProofWindow
+		sectorCost := settings.Prices.RPCAppendSectorsCost(1, expirationHeight-settings.Prices.TipHeight)
+		accountFunding := types.Siacoins(10).Sub(sectorCost.RenterCost())
+		revision := formContractUploadSector(t, types.Siacoins(10), types.Siacoins(20), accountFunding)
+
+		newAllowance := types.Siacoins(10)
```
ChrisSchinnerl, line 445: (nit) I'm not sure I understand "no renter funding" correctly but assuming you mean that the allowance of the renter stays the same, this should probably be `newAllowance := revision.Revision.RemainingAllowance()`?

ChrisSchinnerl (approved):

internal/integration/rhp/v4/rhp4_test.go
n8maninger: It means the renter does not need to fund the transaction. It's copied from coreutils, so I'm not terribly concerned.

PR #801, opened 2025-08-06
Fix panic when failing to create revision txnset

n8maninger:
image

Fixes a panic when broadcasting file contract revisions and another potential race where the transaction set would fail update because the state element proofs were not guaranteed to have the same basis.

Fixes are being released in a hotfix https://github.com/SiaFoundation/hostd/tree/hotfix/v2.3.7

### DIFF
--- a/host/contracts/update.go
+++ b/host/contracts/update.go
@@ -227,13 +227,11 @@ func (cm *Manager) ProcessActions(index types.ChainIndex) error {
 		}
 		cm.wallet.SignV2Inputs(&revisionTxn, toSign)
 
-		basis, revisionTxnSet, err := cm.chain.V2TransactionSet(basis, revisionTxn)
-		if err != nil {
+		if basis, revisionTxnSet, err := cm.chain.V2TransactionSet(basis, revisionTxn); err != nil {
 			cm.wallet.ReleaseInputs(nil, []types.V2Transaction{revisionTxn})
 			log.Error("failed to create transaction set", zap.Error(err))
-		}
-		if err := cm.wallet.BroadcastV2TransactionSet(basis, revisionTxnSet); err != nil {
-			cm.wallet.ReleaseInputs(nil, revisionTxnSet[len(revisionTxnSet)-1:])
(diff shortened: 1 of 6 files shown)


persist/sqlite/contracts.go
```diff
@@ -723,11 +717,38 @@ func proofV2Contracts(tx *txn, index types.ChainIndex) (elements []types.V2FileC
...
+	for i := range elements {
+		proofHeight := elements[i].V2FileContractElement.V2FileContract.ProofHeight
+		err := stmt.QueryRow(proofHeight).Scan(
+			decode(&elements[i].ChainIndexElement.ChainIndex.ID),
+			&elements[i].ChainIndexElement.ChainIndex.Height,
+			decode(&elements[i].ChainIndexElement.StateElement.LeafIndex),
+			decode(&elements[i].ChainIndexElement.StateElement.MerkleProof),
+		)
+		if errors.Is(err, sql.ErrNoRows) {
+			continue // should not happen, but safer to ignore the error than fail every contract
```
ChrisSchinnerl, line 745: If it shouldn't happen, should we at least log it?

host/contracts/update.go
```diff
@@ -181,6 +151,132 @@ func (cm *Manager) buildV2StorageProof(cs consensus.State, fce types.V2FileContr
...
+		return fmt.Errorf("failed to get transaction set parents: %w", err)
+	} else if err := cm.wallet.BroadcastV2TransactionSet(basis, revisionTxnSet); err != nil {
+		cm.wallet.ReleaseInputs(nil, []types.V2Transaction{revisionTxn})
+		return fmt.Errorf("failed to broadcast transaction set: %w", err)
+	}
+	log.Debug("broadcast revision transaction", zap.Stringer("transactionID", revisionTxn.ID()))
+	return nil
+}
+
+func (cm *Manager) broadcastV2StorageProof(cs consensus.State, proofBasis types.ChainIndex, fce types.V2FileContractElement, pi types.ChainIndexElement, log *zap.Logger) error {
```
ChrisSchinnerl, line 191: (nit) instead of `fce` and `pi` you could pass a single `V2ProofElement`.

ChrisSchinnerl (approved):

PR #807, opened 2025-08-11
Fix rejected contract failures

### LINKED ISSUE #806: Rejected Renew Contract Failures
### Current Behavior

When a renewed or refresh contract gets rejected, it causes the previous contract to fail. This isn't new behavior, but renewals are more frequent in V2. 

The cause of the failure is the transfer of sector roots from the old contract to the new contract. When the contract is renewed, we immediately UPDATE the foreign key of the sector roots to the new contract in the database. That means the old contract has no sector roots. If the new contract gets rejected, the old contracts roots cannot be recovered because the new contract may have been revised in the interim.

n8maninger:
Renewed contracts that do not make it on chain will no longer fail and adds a 6 block buffer before storage is reclaimed to ensure small reorgs do not cause unnecessary contract failures. When a contract is renewed, this will temporarily double its contract sectors metric. 

This does slow down RPCRenew and RPCRefresh for larger contracts, but the tradeoff in consistency is worth it.

*Assuming 10TiB contract*
**Old:** ~1.8s
**New:** 2.6s

Fixes #806


ChrisSchinnerl (dismissed):
left a few nits but looks good to me

ChrisSchinnerl (changes requested):
Tests are failing now

PR #809, opened 2025-08-14
Retry rejected contracts

n8mgr:
After #807 when a renewed contract is rejected, the original contract can still complete successfully. However, the renter cannot attempt to renew the contract again or use the active contract. This causes unnecessary churn on renters as they need to repair the data instead of renewing it at no fault of the host.

This PR enables renters to attempt the renewal again as long as the original contract is still active.

### DIFF
--- a/host/contracts/manager_test.go
+++ b/host/contracts/manager_test.go
@@ -241,7 +241,26 @@ func TestV2ContractLifecycle(t *testing.T) {
 			}
 	}
 
-	appendSector := func(t *testing.T, contractID types.FileContractID, fc types.V2FileContract, corrupt bool) (types.V2FileContract, []types.Hash256, proto4.Usage) {
+	fundAccount := func(t *testing.T, contractID types.FileContractID, fc types.V2FileContract, account proto4.Account, amount types.Currency) (types.V2FileContract, proto4.Usage) {
+		fc.RenterOutput.Value = fc.RenterOutput.Value.Sub(amount)
+		fc.HostOutput.Value = fc.HostOutput.Value.Add(amount)
+		fc.RevisionNumber++
+		sigHash := node.Chain.TipState().ContractSigHash(fc)
+		fc.HostSignature = hostKey.SignHash(sigHash)
+		fc.RenterSignature = renterKey.SignHash(sigHash)
+		usage := proto4.Usage{
+			AccountFunding: amount,
+		}
+		_, err := node.Contracts.CreditAccountsWithContract([]proto4.AccountDeposit{
+			{Account: account, Amount: amount},
+		}, contractID, fc, usage)
+		if err != nil {
+			t.Fatal(err)
+		}
+		return fc, usage
+	}
+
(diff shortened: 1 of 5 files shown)


persist/sqlite/contracts.go
```diff
@@ -864,11 +864,52 @@ raw_revision, host_sig, renter_sig, confirmed_revision_number, contract_status,
...
+// If the contract exists, and its contract_status is rejected the
+// existing contract will be deleted so the new contract can be inserted
+// and returns (nil).
+//
+// If the contract exists, and its contract_status is not rejected,
+// [contracts.ErrContractExists] will be returned.
+//
+// If the contract does not exist, it returns (nil).
+
+func v2ContractExists(tx *txn, contractID types.FileContractID) error {
```
ChrisSchinnerl, line 878: `deleteRejectedV2Contract`? The current name doesn't really indicate that a deletion is happening.

persist/sqlite/contracts.go
```diff
@@ -864,11 +864,52 @@ raw_revision, host_sig, renter_sig, confirmed_revision_number, contract_status,
...
+		return nil // contract does not exist
+	} else if err != nil {
+		return fmt.Errorf("failed to check contract existence: %w", err)
+	}
+
+	if status != contracts.V2ContractStatusRejected {
+		return contracts.ErrContractExists
+	}
+
+	_, err = tx.Exec(`DELETE FROM contracts_v2 WHERE id=$1;`, dbID)
```
ChrisSchinnerl, line 892: Did you check that this won't fail due to any of the foreign key constraints? The state elements set the reference to NULL implicitly and `renewed_to` / `renewed_from` too but that still leaves the sector roots and account funding. The roots are probably fine because the sectors get pruned but what about the account funding?

host/contracts/manager_test.go
```diff
@@ -177,6 +175,112 @@ func TestV2ContractLifecycle(t *testing.T) {
...
+		fc.MissedHostValue = fc.MissedHostValue.Sub(collateral)
+		sigHash := node.Chain.TipState().ContractSigHash(fc)
+		fc.HostSignature = hostKey.SignHash(sigHash)
+		fc.RenterSignature = renterKey.SignHash(sigHash)
+
+		usage := proto4.Usage{
+			Storage:          cost,
+			RiskedCollateral: collateral,
+		}
+		err := node.Contracts.ReviseV2Contract(contractID, fc, roots, usage)
```
ChrisSchinnerl, line 277: I'd also create an account and credit it. You can't upload without funding an account first and uploading to temp storage.

ChrisSchinnerl (changes requested):

persist/sqlite/contracts.go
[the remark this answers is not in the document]
n8maninger: You might double check, but a contract that has status “renewed” will have its resolution height field set. 

https://github.com/SiaFoundation/hostd/blob/1f6812a910af70ccfe14aa1e9c2d2f4d35cc77a4/persist/sqlite/contracts.go#L549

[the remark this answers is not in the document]
n8maninger: We also test that here:

https://github.com/SiaFoundation/hostd/blob/78d19807e7a67cba7660db993463846aa37e8573/host/contracts/manager_test.go#L554

PR #814, opened 2025-08-20
refactor: use the built-in max/min to simplify the code

deepdring:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.


n8maninger:
In general, we do not like to accept pure modernization PRs, unless they bring a measurable performance benefit or they are a very clear improvement over the current code.

PR #834, opened 2025-09-22
Archive v1

### LINKED ISSUE #774: V1 contracts in limbo pollute the active contract count
### Current Behavior

In the home page and in the metrics api there is an active contract count. Since the V2 required height there are a number of V1 contracts stuck in limbo and will be forever Active. 

### Expected Behavior

They need to be removed from the metrics from block 530,000 while maintaining consistency of the data, perhaps by flagging them as Completed in the UI.

### Steps to Reproduce

_No response_

### Version

v2.3.5

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 24.04 LTS

### Anything else?

_No response_

n8mgr:
Fixes #774

### DIFF
--- a/persist/sqlite/migrations.go
+++ b/persist/sqlite/migrations.go
@@ -15,7 +15,7 @@ import (
 // migrateVersion45 cleans up unresolved v1 contracts since they will never complete
 // and v1 support has been removed.
 func migrateVersion45(tx *txn, log *zap.Logger) error {
-	rows, err := tx.Query(`SELECT id FROM contracts WHERE status IN ($1, $2)`, contracts.ContractStatusActive, contracts.ContractStatusPending)
+	rows, err := tx.Query(`SELECT id FROM contracts WHERE contract_status IN ($1, $2)`, contracts.ContractStatusActive, contracts.ContractStatusPending)
 	if err != nil {
 		return fmt.Errorf("failed to query active and pending contracts: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


persist/sqlite/migrations.go
```diff
@@ -12,6 +12,73 @@ import (
...
+		if _, err := deleteSectorsStmt.Exec(id); err != nil {
+			return fmt.Errorf("failed to delete contract sector roots: %w", err)
+		} else if _, err := deleteContractAccountFundingStmt.Exec(id); err != nil {
+			return fmt.Errorf("failed to delete contract account funding: %w", err)
+		} else if _, err := deleteContractsStmt.Exec(id); err != nil {
+			return fmt.Errorf("failed to delete contracts: %w", err)
+		}
+	}
+
+	if err := recalcContractAccountFunding(tx, log); err != nil {
```
ChrisSchinnerl, line 68: Do you need to call this? I guess better safe than sorry but at this point the contract and its account funding was deleted and it shouldn't have changed for other contracts?

ChrisSchinnerl (changes requested):
I don't dare approve this as-is because there is too much unrelated but oddly specific stuff happening in this PR. Looks like it could just be refactoring and more thorough testing but in case it's not it would be good to know what triggered it.

The archival migration looks good though.

PR #861, opened 2025-11-07
Fix RenewV2Contract failing for contracts with higher capacity than file size

ChrisSchinnerl:
While investigating why some contracts on the indexer failed with an internal error from the host and then continued to fail with an error indicating that the contract was renewed, I searched Atlas' logs for failed renewals.

There I found this:
```
hostd-mainnet-1  | 2025-11-06T04:55:34Z	ERROR	rhp4.siamux	RPC failed	{"peerAddress": "<redacted>", "streamID": "0e360ba1", "rpc": "RenewContract", "error": "failed to add contract: renewal contract must have same capacity as existing contract", "elapsed": "77.730847ms"}
```
Which also happened to be one of our Zeus contracts.

Turns out the sanity check in `RenewV2Contract` prevented contracts with a capacity greater than its file size from being renewed. Which happens after broadcasting the renewed contract to the network. Leading to these contracts no longer being able to renew since they were.

### DIFF
--- a/host/contracts/manager.go
+++ b/host/contracts/manager.go
@@ -294,21 +294,17 @@ func (cm *Manager) RenewV2Contract(renewal rhp4.TransactionSet, usage proto4.Usa
 	}
 	defer done()
 
-	badRequest := func(format string, args ...any) error {
-		return proto4.NewRPCError(proto4.ErrorCodeBadRequest, fmt.Sprintf(format, args...))
-	}
-
 	renewalSet := renewal.Transactions
 	if len(renewalSet) == 0 {
-		return badRequest("no renewal transactions provided")
+		return errors.New("no renewal transactions provided")
 	} else if len(renewalSet[len(renewalSet)-1].FileContractResolutions) != 1 {
-		return badRequest("last transaction must contain one file contract resolution")
+		return errors.New("last transaction must contain one file contract resolution")
 	}
 
 	resolutionTxn := renewalSet[len(renewalSet)-1]
 	resolution, ok := resolutionTxn.FileContractResolutions[0].Resolution.(*types.V2FileContractRenewal)
 	if !ok {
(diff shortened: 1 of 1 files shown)


config/compat.go
```diff
@@ -97,6 +97,7 @@ type configV112 struct {
 	} `yaml:"rhp3,omitempty"`
 	Log struct {
 		// Path is the directory to store the hostd.log file.
+		//
```
ChrisSchinnerl: Inb4 anyone complains about this madness. It's the updated linter that wants this now. Including getting rid of the `+build` tags.

host/contracts/manager.go
```diff
@@ -288,17 +288,21 @@ func (cm *Manager) RenewV2Contract(renewal rhp4.TransactionSet, usage proto4.Usa
 	}
 	defer done()
 
+	badRequest := func(format string, args ...any) error {
```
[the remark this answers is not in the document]
n8mgr: I'd remove it. These validations are supposed to be handled in `coreutils`. Realistically, they shouldn't even be here, but it was a bit of defensive programming that bit us in the ass.

host/contracts/manager.go
```diff
@@ -309,18 +313,21 @@ func (cm *Manager) RenewV2Contract(renewal rhp4.TransactionSet, usage proto4.Usa
 	fc := resolution.NewContract
 
 	// sanity checks
+	refresh := fc.ExpirationHeight == existing.ExpirationHeight
```
n8mgr: I'd get rid of all of these and handle it in coreutils. The rhp4 server should be handling these kinds of validations.

n8mgr (changes requested):
Make sure `coreutils` is actually validating the renewal correctly now that the sanity checks are removed.

PR #880, opened 2025-11-26
Make max collateral a multiplier

### LINKED ISSUE #857: Make max collateral a multiplier
### Discussed in https://github.com/SiaFoundation/hostd/discussions/823

<div type='discussions-op-text'>

<sup>Originally posted by **NickJH** September  5, 2025</sup>
### Description

As the recommendation is to set max collateral to `storage price * collateral multiplier * 10`, rather than have the user calculate it on any price change and remember to update it, use a max collateral multiplier like the collateral multiplier.

See https://discord.com/channels/809849352516141067/946327219708514344/1413580892781154395.

### Version

v2.4.0

### What operating system are you running (e.g. Ubuntu 22.04, macOS, Windows 11)?

Ubunto 24.04 LTS but it is irrelevant

### Anything else?

_No response_</div>

n8mgr:
This version is draft because it will currently fail RHP4 tests because it treats max collateral as a TB/byte/block like storage price and collateral instead of as a fixed absolute value. If we want to continue down this path, we will need to similarly change the RHP4 validation and bump the protocol version.

Alternatively, we can calculate the actual max collateral value as described in the issue and assume `max contract duration`.

Fixes #857

### DIFF
--- a/host/settings/settings.go
+++ b/host/settings/settings.go
@@ -103,8 +103,8 @@ type (
 		BaseRPCPrice      types.Currency `json:"baseRPCPrice"`
 		SectorAccessPrice types.Currency `json:"sectorAccessPrice"`
 
-		CollateralMultiplier float64        `json:"collateralMultiplier"`
-		MaxCollateral        types.Currency `json:"maxCollateral"`
+		CollateralMultiplier    float64 `json:"collateralMultiplier"`
+		MaxCollateralMultiplier float64 `json:"maxCollateralMultiplier"`
 
 		StoragePrice types.Currency `json:"storagePrice"`
 		EgressPrice  types.Currency `json:"egressPrice"`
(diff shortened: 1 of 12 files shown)


host/settings/settings.go
```diff
@@ -175,8 +175,8 @@ var (
 		BaseRPCPrice:      types.Siacoins(1).Div64(1e6), // 1 SC / million RPCs
 		SectorAccessPrice: types.Siacoins(1).Div64(1e6), // 1 SC / million sectors
 
-		CollateralMultiplier: 2.0, // 2x storage price
-		MaxCollateral:        types.Siacoins(1000),
+		CollateralMultiplier:    2.0,  // 2x storage price
+		MaxCollateralMultiplier: 10.0, // 10x collateral
```
ChrisSchinnerl: Doesn't making this a multiplier effectively remove the max? I'm not sure how you'd apply it in the validation code.

e.g. a host offers 1H / byte / month then the max becomes 10H / byte / month. But what does that even mean? 

In my head that would only work if the multiplier was not just a multiplier but the actual storage you expect a renter to be able to store in a contract. Like `10 * 1<<40` and then the max is 10TB of collateral / month.

PR #887, opened 2025-12-02
Support syncing from a checkpoint

n8mgr:


cmd/hostd/run.go
```diff
@@ -232,22 +237,84 @@ func runRootCmd(ctx context.Context, cfg config.Config, walletKey types.PrivateK
...
+		if err != nil {
+			return fmt.Errorf("failed to retrieve checkpoint: %w", err)
+		}
+		log.Debug("retrieved checkpoint")
+
+		if err := store.SetCheckpoint(checkpoint); err != nil {
+			return fmt.Errorf("failed to set checkpoint in store: %w", err)
+		}
+
+		bdb, err := coreutils.OpenBoltChainDB(filepath.Join(cfg.Directory, "consensus.db"))
```
peterjan: `nit` you need this in both branches, if you move it up the if-else is simplified

peterjan (approved):

ChrisSchinnerl (dismissed):
Lgtm. Would be nice if you could run it once more against Zen for good measure.

PR #951, opened 2026-04-08
Update configs in README.md

CtrlAltDefeat94:
- Updates the example config to use new default values. 
- Adds a table with all settings in similar style as Renterd has. 
- Removes deprecated fields such as RHP2 etc.

Closes #937

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ log:
 | `Syncer.Peers` | Additional peer addresses to add to the peer store at startup | `[]` | - | - | `syncer.peers[]` |
 | `Consensus.Network` | Consensus network to join | `mainnet` | `--network` | - | `consensus.network` |
 | `Consensus.IndexBatchSize` | Batch size used by the index manager while processing updates | `1000` | - | - | `consensus.indexBatchSize` |
-| `Consensus.PruneTarget` | Number of recent blocks to retain when pruning consensus data; `0` disables pruning | `0` | - | - | `consensus.pruneTarget` |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,39 +107,49 @@ log:
...
--rhp2 string
-	address to listen on for RHP2 connections (default ":9982")
--rhp3 string
-	address to listen on for TCP RHP3 connections (default ":9983")
--rhp4 string
-        address to listen on for RHP4 connections
--env
-	disable stdin prompts for environment variables (default false)
-```
+### Configuration Reference
```
peterjan, line 110: I would personally get rid of the table. I think the README should definitely mention environment variables and whatever config is required to run a host in its default setup. I've always regretted adding that for `renterd`. Considering you did all this work though, probably good to wait for Nate's input on this one.

CtrlAltDefeat94: IMO it's not a good idea to stick to an undesirable format just because of sunk cost fallacy.
The annoying part was getting all information together, if you have a different idea on how to display the information I'm open to updating it.

peterjan (approved):

README.md
peterjan: Fair point. To make a suggestion though I would list environment variables and the most important CLI flags like http.address, network, instant syncing. For other, more obscure knobs/config I would refer to `-- help` and make sure the help command has good description and indication of default values.

ChrisSchinnerl (approved):

PR #955, opened 2026-04-21
Fix failing refresh after wallet address change

ChrisSchinnerl:
Related to https://github.com/SiaFoundation/coreutils/pull/418


n8mgr (changes requested):
I'd add a defensive test that the renter can't pass any old address

PR #959, opened 2026-05-10
add support for RHP4 pools

n8mgr:


persist/sqlite/pools.go
```diff
@@ -0,0 +1,241 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to prepare get pool statement: %w", err)
+		}
+		defer getPoolStmt.Close()
+
+		poolIDs := make([]int64, len(attachments))
+		for i, a := range attachments {
+			err := getPoolStmt.QueryRow(encode(a.Pool)).Scan(&poolIDs[i])
+			if errors.Is(err, sql.ErrNoRows) {
+				return proto4.ErrPoolNotFound
```
ChrisSchinnerl, line 118: Probably not a big deal but I want to highlight that a single unknown pool fails the whole RPC. Mostly so that when we talk about that in a year and how annoying it is we can look back at this comment and be slightly more annoyed.

ChrisSchinnerl (dismissed):

persist/sqlite/pools.go
peterjan: haha 😂

n8mgr: I will abstain from fixing it so we can revel in that joy this time next year

persist/sqlite/accounts.go
```diff
@@ -57,29 +57,63 @@ func (s *Store) RHP4AccountBalance(account proto4.Account) (balance types.Curren
...
+		}
+
+		// try all attached pools
+		for _, p := range attached {
+			if cost.IsZero() {
+				break
+			}
+			poolUsage := takeRHP4Usage(&usage, p.Balance)
+			if taken := poolUsage.RenterCost(); !taken.IsZero() {
+				if _, err := tx.Exec(`UPDATE rhp4_pools SET balance=$1 WHERE id=$2`, encode(p.Balance.Sub(taken)), p.ID); err != nil {
```
chris124567: why not use prepared statements

persist/sqlite/init.sql
```diff
@@ -241,6 +241,30 @@ CREATE TABLE contract_v2_account_funding (
...
+CREATE TABLE rhp4_account_pool_attachments (
+	id INTEGER PRIMARY KEY,
+	account_id INTEGER NOT NULL REFERENCES accounts(id),
+	pool_id INTEGER NOT NULL REFERENCES rhp4_pools(id),
+	UNIQUE (account_id, pool_id)
+);
+CREATE INDEX rhp4_account_pool_attachments_account_id_id ON rhp4_account_pool_attachments(account_id, id);
+CREATE INDEX rhp4_account_pool_attachments_pool_id ON rhp4_account_pool_attachments(pool_id);
+
 CREATE TABLE host_stats (
```
chris124567, line 268: Maybe we should update metricAccountBalance and metricActiveAccounts or create new pool metrics

persist/sqlite/init.sql
```diff
@@ -241,6 +241,30 @@ CREATE TABLE contract_v2_account_funding (
 	UNIQUE (contract_id, account_id)
 );
 
+CREATE TABLE rhp4_pools (
+	id INTEGER PRIMARY KEY,
+	pool_id BLOB UNIQUE NOT NULL,
+	balance BLOB NOT NULL
+);
+
+CREATE TABLE contract_v2_pool_funding (
```
chris124567, line 250: Update resetRejectedContractAccountFunding to handle this too

persist/sqlite/init.sql
n8mgr: Maybe in a follow up. Pretty sure no one is looking at the account metrics atm anyway.

persist/sqlite/init.sql
```diff
@@ -241,6 +241,30 @@ CREATE TABLE contract_v2_account_funding (
...
+	contract_id INTEGER NOT NULL REFERENCES contracts_v2(id),
+	pool_id INTEGER NOT NULL REFERENCES rhp4_pools(id),
+	amount BLOB NOT NULL,
+	UNIQUE (contract_id, pool_id)
+);
+CREATE INDEX contract_v2_pool_funding_pool_id ON contract_v2_pool_funding(pool_id);
+
+CREATE TABLE rhp4_account_pool_attachments (
+	id INTEGER PRIMARY KEY,
+	account_id INTEGER NOT NULL REFERENCES accounts(id),
```
peterjan, line 261: There's no cascade delete here so deleting accounts will be impossible if it's referenced by an attachment  here. Looking into this however I noticed that `PruneAccounts` is actually never used, it can not be because it checks on `expiration_height` which isn't even a column on the accounts table, it should `expiration_timestamp.`

persist/sqlite/consensus.go
```diff
@@ -1972,6 +1986,64 @@ func resetRejectedContractAccountFunding(tx *txn, contractDBID int64, log *zap.L
 	return nil
 }
 
+// resetRejectedPoolFunding rolls back pool deposits funded by a rejected
+// contract. Pool rows are shared across attached accounts and other funding
+// contracts, so the balance is reduced in place rather than the row deleted.
+func resetRejectedPoolFunding(tx *txn, contractDBID int64, log *zap.Logger) error {
```
peterjan, line 1982: This is an important piece of logic that's untested.

peterjan (changes requested):

chris124567 (commented):
Courtesy of AI but there may be issues with this  https://github.com/SiaFoundation/hostd/blob/9ac470356175ee76fdd92fac7a991b644d0d1072/persist/sqlite/consensus.go#L1968 if the account is still referenced in rhp4_account_pool_attachments

Edit: I think this is what PJ was referring to above

chris124567 (approved):

ChrisSchinnerl (approved):

peterjan (approved):

PR #965, opened 2026-06-04
Fix alert

### LINKED ISSUE #487: Problem with hostd critical alert reporting
### Current Behavior

Right now I just had an issue (not sure how or why it happened, it might have been a resize as that is what I was doing at the time).  Where my the last volume I was using which I just added space on a raid volume was locked.  I never locked it so I am thinking it might have happened during a resize.  Anyways all the other volumes filled up and although there is a ton of space on the server there was no ability to write stuff to the hosts system.

Image

Despite no ability to write to the system, there were no errors logged in the alerts system.  it did however show up on the troubleshooter (which it should have, it should also have shown up in hostd).

Image

This indicates there is a problem with the alerts system which needs to be fixed

### Expected Behavior

The alert should have shown up in hosted alerts system (bell).  The problem went unfixed for 6 days (and a lot of contracts lost) due to hosted indicating everything was ok when in fact it was not.

### Steps to Reproduce

_No response_

### Version

v1.1.2

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 22.04

### Anything else?

_No response_

chris124567:
Close #487


host/storage/storage.go
```diff
@@ -707,6 +794,7 @@ func (vm *VolumeManager) ResizeVolume(ctx context.Context, id int64, maxSectors
 			return fmt.Errorf("failed to set volume %v to read-only: %w", id, err)
 		}
 		resetReadOnly = true
+		vm.updateNoWritableStorageAlert(nil)
 	}
```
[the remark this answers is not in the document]
chris124567: fixed

host/storage/storage.go
```diff
@@ -687,9 +774,8 @@ func (vm *VolumeManager) ResizeVolume(ctx context.Context, id int64, maxSectors
...
 
 	vm.mu.Lock()
-	defer vm.mu.Unlock()
```
peterjan, line 690: This looked a bit fishy to me at first but it's probably fine. The locking here is a little odd though, for example: what happens if this thread sets read only to true, but then the host gets shut down right after, wouldn't `tg.AddContext` fail and read only never gets reset? cc @n8mgr

peterjan (commented):
I think it's fine but Nate would have a better idea of course. I would personally consider re-evaluating the alert on an interval as opposed to explicitly calling `updateNoWritableStorageAlert` because it's probably fine to have a stale alert for one minute or so and it's quite annoying/dirty to have to remember the alert in every action.

host/storage/storage.go
```diff
@@ -361,6 +368,80 @@ func (vm *VolumeManager) writeSector(root types.Hash256, data *[proto4.SectorSiz
 		vm.cache.Add(root, data)
 		return vol.Sync()
 	})
+	if errors.Is(err, ErrNotEnoughStorage) {
+		vm.updateNoWritableStorageAlert(err)
+	} else if err == nil && vm.alerts.IsActive(NoWritableStorageAlertID) {
+		// a nil error does not prove writable space remains: the sector may
```
ChrisSchinnerl, line 374: If a nil error doesn't prove that space remains why does a nil error + a registered alert prove it?

chris124567: Well if we have a successful write then we want to re-evaluate whether the alert should stay on.  But there's no sense in re-evaluating if the write succeeded and there's currently no alert registered.  If the write succeeds and the alert is still registered, then we should see if we can dismiss the alert.

host/storage/storage.go
```diff
@@ -25,6 +25,10 @@ const (
 	MaxTempSectorBlocks = 144 * 7 // 7 days
 )
 
+// NoWritableStorageAlertID is the ID of the alert registered when volumes
+// have free space, but none of it is writable.
+var NoWritableStorageAlertID = types.HashBytes([]byte("storage:noWritableStorage"))
```
ChrisSchinnerl: Why not `frand.Entropy256()` like we do for all the other ids?

host/storage/storage.go
ChrisSchinnerl: I would also like to know why this was updated. Because in other places like `loadVolumes` we still hold the lock while calling `vm.updateNoWritableStorageAlert`.

chris124567: I added the defer back for readability reasons but I don't think it's necessary.  I just noticed that we could reduce the amount of time the lock was held.  We don't actually need the lock during `updateNoWritableStorageAlert`.

ChrisSchinnerl (dismissed):

peterjan (dismissed):
I'm fine approving but this question is still open for me https://github.com/SiaFoundation/hostd/pull/965#discussion_r3361167100 - that was already in the code though so I guess it's fine but still

host/storage/storage.go
chris124567: Ok I think I fixed that issue even though it was preexisting

peterjan (approved):

ChrisSchinnerl (approved):

PR #966, opened 2026-06-05
Repair v2 contract statuses in database

### LINKED ISSUE #912: Something wrong with contract statuses?
### Current Behavior

Please see this screenshot:

Image
The bottom two contracts say active, finishing on 18/01, paying out on 20/01. The both have been renewed to other contracts. Shouldn't they be Complete or Renewed?

Contracts are `24f9ede7de6b61ddbc74aa8ea88cdff49cdef64f3fa4348c76b861fbad58f874` and `405b75eab11aec2a4adf7cf874e63a7d86434d011700e9132c3f356e93610bea`.

### Expected Behavior

If the contracts have completed, they should show Complete. If they have been renewed, they should show Renewed. I don't see any way a renewed contract should be showing Active.

### Steps to Reproduce

_No response_

### Version

v2.5.1

### Anything else?

_No response_

peterjan:
This PR repairs v2 contracts left in active after renewal. There was a bug where a single block could contain both a revision and the renewal resolution for the same contract: the renewal was dropped and the parent's contract_status, resolution_block_id, and resolution_height were never written. The forward fix landed in #930, this PR adds a migration so existing deployments get corrected without a chain rescan. The historical time series in host_stats is not backfilled. That would require a chain rescan.

Side benefit: ExpireV2ContractSectors is gated on resolution_height IS NOT NULL, so the stuck rows previously kept their sector roots around forever. They get cleaned up on the next call.

Closes #912

### DIFF
--- a/persist/sqlite/migrations.go
+++ b/persist/sqlite/migrations.go
@@ -13,6 +13,64 @@ import (
 	"go.uber.org/zap"
 )
 
+// migrateVersion51 corrects v2 contracts left in active status when their
+// renewal landed in the same block as a revision.
+func migrateVersion51(tx *txn, log *zap.Logger) error {
+	// fetch active contracts with a renewal that has been confirmed
+	// on chain but the old contract was not marked as renewed
+	rows, err := tx.Query(`SELECT c.id, cc.confirmation_index
+FROM contracts_v2 c
+INNER JOIN contracts_v2 cc ON c.renewed_to = cc.id
+WHERE c.contract_status = ?
+AND cc.confirmation_index IS NOT NULL`, contracts.V2ContractStatusActive)
+	if err != nil {
+		return fmt.Errorf("failed to query stuck contracts: %w", err)
+	}
+	defer rows.Close()
+
+	affected := make(map[int64]types.ChainIndex, 0)
+	for rows.Next() {
+		var dbID int64
+		var ci types.ChainIndex
+		if err := rows.Scan(&dbID, decode(&ci)); err != nil {
+			return fmt.Errorf("failed to scan stuck contract: %w", err)
+		}
+		affected[dbID] = ci
+	}
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("failed to iterate stuck contracts: %w", err)
+	} else if err := rows.Close(); err != nil {
+		return fmt.Errorf("failed to close stuck contracts rows: %w", err)
+	} else if len(affected) == 0 {
+		return nil
+	}
+
+	// update contract status to renewed for the affected contracts and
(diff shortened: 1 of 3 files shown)


persist/sqlite/migrations.go
```diff
@@ -13,6 +13,64 @@ import (
...
+	} else if err := rows.Close(); err != nil {
+		return fmt.Errorf("failed to close stuck contracts rows: %w", err)
+	} else if len(affected) == 0 {
+		return nil
+	}
+
+	// update contract status to renewed for the affected contracts and
+	// set the resolution index to the confirmation index of the renewal
+	log.Info("repairing v2 contracts left active after renewal", zap.Int("count", len(affected)))
+	updateStmt, err := tx.Prepare(`UPDATE contracts_v2 SET contract_status=?, resolution_block_id=?, resolution_height=? WHERE id=?`)
```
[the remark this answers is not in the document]
peterjan: I don't think this is an issue because that database state is not possible

persist/sqlite/migrations.go
```diff
@@ -13,6 +13,64 @@ import (
...
+	defer updateStmt.Close()
+
+	for dbID, ci := range affected {
+		if _, err := updateStmt.Exec(contracts.V2ContractStatusRenewed, encode(ci.ID), ci.Height, dbID); err != nil {
+			return fmt.Errorf("failed to repair contract %d: %w", dbID, err)
+		}
+	}
+
+	// historical time series points are not backfilled
+	if err := recalcContractCountMetrics(tx); err != nil {
```
ChrisSchinnerl, line 64: This is a pretty massive migration for just fixing a hand full of contracts but getting the logic right to only update the deltas might be a bit of a nightmare.

persist/sqlite/migrations_test.go
```diff
@@ -536,3 +536,153 @@ VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $
...
+		t.Fatal("unexpected", c.ResolutionIndex)
+	}
+
+	// assert lone active contract unchanged
+	c, err = store.V2Contract(loneActive)
+	if err != nil {
+		t.Fatal(err)
+	} else if c.Status != contracts.V2ContractStatusActive {
+		t.Fatal("unexpected", c.Status)
+	}
```
ChrisSchinnerl, line 689: This test doesn't verify the probably most important part of the migration which is the metrics being corrected.

ChrisSchinnerl (changes requested):

n8mgr (approved):

ChrisSchinnerl (approved):

PR #975, opened 2026-07-08
Add postgres Store

ChrisSchinnerl:
This PR is the first of probably quite a few to introduce a Postgres driver to hostd.
It starts by adding a `init.sql` which is similar to SQLite's but with Postgres types as well as similar boilerplate code we use in `indexd`. The only difference between `indexd` and `hostd` is that this PR uses an embedded Postgres db for testing which makes running tests more convenient. 

Since we have 100+ store interface methods the goal is to follow up with at least 1 PR per store interface in the host but considering some of those are quite large probably a lot more than that. 
I really want to avoid this being a dumb translation and instead make sure it's making good use of Postgres' capabilities.


persist/postgres/store.go
```diff
@@ -0,0 +1,220 @@
...
+
+	db, err := sql.Open("pgx", ci.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to open database: %w", err)
+	} else if err := db.PingContext(ctx); err != nil {
+		db.Close()
+		return nil, fmt.Errorf("failed to connect to database: %w", err)
+	}
+
+	store := &Store{
```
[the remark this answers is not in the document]
ChrisSchinnerl: I don't think that's right. The pgx driver should set a sane limit depending on CPU count.

persist/postgres/init.sql
```diff
@@ -0,0 +1,351 @@
...
+
+CREATE TABLE volume_sectors (
+	id BIGSERIAL PRIMARY KEY,
+	volume_id BIGINT NOT NULL REFERENCES storage_volumes (id), -- all sectors will need to be migrated first when deleting a volume
+	volume_index BIGINT NOT NULL,
+	sector_id BIGINT UNIQUE REFERENCES stored_sectors (id),
+	sector_writes BIGINT NOT NULL DEFAULT 0,
+	UNIQUE (volume_id, volume_index)
+);
+CREATE INDEX volume_sectors_sector_writes_volume_id_sector_id_volume_index_compound ON volume_sectors(sector_writes ASC, volume_id, sector_id, volume_index) WHERE sector_id IS NULL;
```
peterjan, line 61: do we care this index name exceeds the 63 char limit postgres imposes? (there's 3 in total)

persist/postgres/sql.go
```diff
@@ -0,0 +1,116 @@
...
+	r := tx.Tx.QueryRow(ctx, query, args...)
+	if dur := time.Since(start); dur > longQueryDuration {
+		tx.log.Debug("slow query row", zap.String("query", query), zap.Duration("elapsed", dur), zap.Stack("stack"))
+	}
+	return &row{r, tx.log.Named("row")}
+}
+
+// getDBVersion returns the current version of the database.
+func getDBVersion(ctx context.Context, pool *pgxpool.Pool) (version int64) {
+	// error is ignored -- the database may not have been initialized yet.
```
peterjan, line 106: this is kind of odd? we do this for s3d and indexd too but wouldn't it be better to check explicitly and return an error if it's not initialized? or perhaps even retry with a small backoff and error out after a second or so

persist/postgres/init.sql
```diff
@@ -0,0 +1,351 @@
...
+CREATE INDEX wallet_events_chain_index ON wallet_events(chain_index);
+CREATE INDEX wallet_events_maturity_height ON wallet_events(maturity_height DESC);
+
+CREATE TABLE stored_sectors (
+	id BIGSERIAL PRIMARY KEY,
+	sector_root BYTEA UNIQUE NOT NULL CHECK (LENGTH(sector_root) = 32),
+	cached_subtree_roots BYTEA,
+	last_access_timestamp TIMESTAMP WITH TIME ZONE NOT NULL
+);
+CREATE INDEX stored_sectors_sector_root ON stored_sectors(sector_root);
```
chris124567, line 39: We don't need an index here because we already have `UNIQUE` on that column.  It seems there are a few more places this happens

PR #991, opened 2026-09-03
Fix high CPU usage when pruning sectors on host that doesn't have any sectors to prune

ChrisSchinnerl:
Fixes an issue where some hosts experience high cpu usage when they have no sectors to prune since the query performs a full table scan.

Added another benchmark and updated the existing one to set cached subtrees to extend the row length to what it would be in production.

Some results:

| benchmark | before | after | delta |
|---|---:|---:|---:|
| `PruneSectors/retained=10` | 1.075 ms | 0.796 ms | −25.9% |
| `PruneSectors/retained=100` | 1.092 ms | 0.804 ms | −26.4% |
| `PruneSectors/retained=1000` | 1.266 ms | 0.975 ms | −23.0% |
| `PruneSectors/retained=25600` | 1.388 ms | 1.105 ms | −20.4% |
| `PruneSectorsFullScan/sectors=1000` | 0.282 ms | 0.165 ms | −41.5% |
| `PruneSectorsFullScan/sectors=10000` | 3.662 ms | 1.476 ms | −59.7% |
| `PruneSectorsFullScan/sectors=25600` | 28.861 ms | 3.937 ms | −86.4% |


persist/sqlite/init.sql
```diff
@@ -38,6 +38,7 @@ CREATE TABLE stored_sectors (
 );
 CREATE INDEX stored_sectors_sector_root ON stored_sectors(sector_root);
 CREATE INDEX stored_sectors_last_access ON stored_sectors(last_access_timestamp);
```
chris124567: I don't think we need  `stored_sectors_last_access` anymore

n8mgr (approved):

PR #993, opened 2026-09-04
Move sector cache to dedicated table

ChrisSchinnerl:
I have been digging a bit more since I didn't get why not everyone was affected by the recent CPU spike. So here is what I think happened.

SQLite is Index-Organized while Postgres is Heap based. So in SQLite the whole table data is stored in a B-Tree clustered by rowid and a primary key just maps to a rowid. In Postgres the primary key is stored separately in its own B-Tree.

So for the stored_sectors table that means joining on the id and filtering by it scans a B-Tree that holds very few rows per page due to the cache. Not the full 32kib but probably around half a kib. The reason being some [math](https://www.sqlite.org/fileformat.html#b_tree_pages) that happens on the page size which determines the minimum amount of data actually kept inline before spilling over into the overflow pages.

The fix we implemented, which I still think was the right hotfix, was bypassing the B-Tree by giving the query a covering index. 

Here comes the interesting part. When you installed hostd before we added the cache, the migration appended the column at the end. Rather than before the last_access_timestamp. This directly affects the scan time because a row is decoded in column order, and anything past the inlined portion means walking the row's overflow pages one at a time. So if it needs column 1 and 2 it reads only 1 and 2. If it needs 1 and 3 it will read 1, 2 and 3. Meaning that a migrated database performs significantly faster due to much smaller reads.

During some local testing I realized that the same issue also causes other queries to be slower than necessary. But saw some significant improvements after moving the cache to a separate table to avoid reading it on any hot path that doesn't need to read it.

The migration itself should also not be as bad as you might think. Dropping a column should not cause the database to grow on disk and frees the overflow pages immediately. Which are then reused when rebuilding the cache. The only waste is about 1-2% of the cache which remains as sparse space within the `stored_sectors` B-Tree which requires a vacuum. But there is going to be a lot of disk i/o for hosts as they rebuild their caches.

### DIFF
--- a/persist/sqlite/init.sql
+++ b/persist/sqlite/init.sql
@@ -33,11 +33,14 @@ CREATE INDEX wallet_events_maturity_height ON wallet_events(maturity_height DESC
 CREATE TABLE stored_sectors (
 	id INTEGER PRIMARY KEY,
 	sector_root BLOB UNIQUE NOT NULL,
-	cached_subtree_roots BLOB,
 	last_access_timestamp INTEGER NOT NULL
 );
 CREATE INDEX stored_sectors_sector_root ON stored_sectors(sector_root);
-CREATE INDEX stored_sectors_id_last_access ON stored_sectors(id, last_access_timestamp);
+
+CREATE TABLE sector_subtree_cache (
+	sector_id INTEGER PRIMARY KEY REFERENCES stored_sectors(id) ON DELETE CASCADE,
+	subtree_roots BLOB NOT NULL
+);
 
 CREATE TABLE storage_volumes (
 	id INTEGER PRIMARY KEY,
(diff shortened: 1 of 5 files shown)


persist/sqlite/init.sql
```diff
@@ -33,12 +33,16 @@ CREATE INDEX wallet_events_maturity_height ON wallet_events(maturity_height DESC
 CREATE TABLE stored_sectors (
 	id INTEGER PRIMARY KEY,
 	sector_root BLOB UNIQUE NOT NULL,
-	cached_subtree_roots BLOB,
 	last_access_timestamp INTEGER NOT NULL
 );
 CREATE INDEX stored_sectors_sector_root ON stored_sectors(sector_root);
 CREATE INDEX stored_sectors_id_last_access ON stored_sectors(id, last_access_timestamp);
```
ChrisSchinnerl: I'm torn about removing this index. A fresh host doesn't need it anymore but a migrated host might still benefit to avoid the sparse pages.

n8mgr (approved):

PR #994, opened 2026-09-04
Enable concurrent SQLite readers

n8mgr:
Re-enables concurrent SQLite readers and adds a new `writeTransaction` which changes the transaction isolation level for a few hot paths that do a select then read and cause a significant amount of db retries. Also ensures that all methods are retry safe.


persist/sqlite/store.go
```diff
@@ -25,45 +25,31 @@ type (
...
-	for ; attempt < maxRetryAttempts; attempt++ {
-		attemptStart := time.Now()
-		log := log.With(zap.Int("attempt", attempt))
-		err = doTransaction(s.db, log, fn)
-		if err == nil {
-			// no error, break out of the loop
-			return nil
-		}
+	return retryTransaction(s.readerDB, log, fn)
+}
```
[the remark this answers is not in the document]
n8mgr: We can clean the remainder of the paths up later. The big ones are converted.

persist/sqlite/sql.go
peterjan, line 145: We swallow a potential error here if the prepare is slow...,  I found out when trying to reproduce `ErrBusy`... this was already in master but it makes sense to take it along with this PR imo.

persist/sqlite/store.go
```diff
@@ -309,21 +309,30 @@ func ForeignKeyCheck(ctx context.Context, fp string, log *zap.Logger) error {
...
 func OpenDatabase(fp string, log *zap.Logger) (*Store, error) {
-	db, err := sql.Open("sqlite3", sqliteFilepath(fp))
+	readerDB, err := sql.Open("sqlite3", readerFilepath(fp))
 	if err != nil {
 		return nil, fmt.Errorf("failed to open database: %w", err)
 	}
 
-	// set the number of open connections to 1 to prevent "database is locked"
-	// errors
-	db.SetMaxOpenConns(1)
```
peterjan, line 319: Now that the reader DB has unbounded max open conns, do we need to set `MaxIdleConns` and `ConnMaxIdleTime`? MaxIdleConns defaults to 2, which never mattered while max open conns was 1, but now anything returned above the second connection gets closed instead of reused. Seems like it could be a lot of needless churn.

peterjan (dismissed):
Approved the PR but left two comments we might want to F/U with

ChrisSchinnerl (dismissed):
Got nothing to add to PJ's comments

persist/sqlite/store.go
n8mgr: We didn't change it in any of our other software, so I'm unsure what a good value would be. We can consider it once we have some runtime with the defaults.

PR #999, opened 2026-09-09
Use reference counter for pruning instead of lookups

n8mgr:
Switches to a reference counter for sector pruning instead of look-ups and removes the last accessed timestamp. This reduces lockups due to the prune loop and reduces IO during sector reads. Only writes now lock the sector location. This is the same system we used before the v2 changes.

### DIFF
--- a/host/storage/storage.go
+++ b/host/storage/storage.go
@@ -334,47 +334,37 @@ func (vm *VolumeManager) shrinkVolume(ctx context.Context, id int64, volume *vol
 }
 
 // writeSector atomically adds a sector to the database and writes it to disk
-func (vm *VolumeManager) writeSector(root types.Hash256, data *[proto4.SectorSize]byte) error {
-	err := vm.vs.StoreSector(root, func(loc SectorLocation) error {
-		start := time.Now()
+func (vm *VolumeManager) writeSector(loc SectorLocation, root types.Hash256, data *[proto4.SectorSize]byte) error {
+	start := time.Now()
 
-		vm.mu.Lock()
-		vol, ok := vm.volumes[loc.Volume]
-		vm.mu.Unlock()
-		if !ok {
-			return fmt.Errorf("volume %v not found", loc.Volume)
-		}
-
-		// write the sector to the volume
-		if err := vol.WriteSector(data, loc.Index); err != nil {
-			stats := vol.Stats()
-			vm.alerts.Register(alerts.Alert{
-				ID:       vol.alertID("write"),
-				Severity: alerts.SeverityError,
-				Message:  "Failed to write sector",
-				Data: map[string]any{
-					"volume":       vol.Location(),
-					"failedReads":  stats.FailedReads,
-					"failedWrites": stats.FailedWrites,
-					"sector":       root,
-					"error":        err.Error(),
-				},
-				Timestamp: time.Now(),
-			})
-			return err
-		}
-		vm.log.Debug("wrote sector", zap.String("root", root.String()), zap.Int64("volume", loc.Volume), zap.Uint64("index", loc.Index), zap.Duration("elapsed", time.Since(start)))
-
-		return vol.Sync()
-	})
-	if errors.Is(err, ErrNotEnoughStorage) {
-		vm.updateNoWritableStorageAlert(err)
-	} else if err == nil && vm.alerts.IsActive(NoWritableStorageAlertID) {
-		// a nil error does not prove writable space remains: the sector may
-		// already have been stored, so re-evaluate instead of dismissing
-		vm.updateNoWritableStorageAlert(nil)
+	vm.mu.Lock()
+	vol, ok := vm.volumes[loc.Volume]
+	vm.mu.Unlock()
+	if !ok {
+		return fmt.Errorf("volume %v not found", loc.Volume)
 	}
-	return err
+
+	// write the sector to the volume
+	if err := vol.WriteSector(data, loc.Index); err != nil {
+		stats := vol.Stats()
+		vm.alerts.Register(alerts.Alert{
+			ID:       vol.alertID("write"),
+			Severity: alerts.SeverityError,
+			Message:  "Failed to write sector",
+			Data: map[string]any{
+				"volume":       vol.Location(),
+				"failedReads":  stats.FailedReads,
+				"failedWrites": stats.FailedWrites,
+				"sector":       root,
+				"error":        err.Error(),
+			},
+			Timestamp: time.Now(),
+		})
+		return err
+	}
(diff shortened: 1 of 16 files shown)


host/storage/storage.go
```diff
@@ -340,49 +340,37 @@ func (vm *VolumeManager) shrinkVolume(ctx context.Context, id int64, volume *vol
...
+			},
+			Timestamp: time.Now(),
+		})
+		return err
 	}
-	return err
+	vm.log.Debug("wrote sector", zap.Stringer("root", root), zap.Int64("volume", loc.Volume), zap.Uint64("index", loc.Index), zap.Duration("elapsed", time.Since(start)))
+
+	return vol.Sync()
 }
```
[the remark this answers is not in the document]
n8mgr: The sector root comes from a trusted source. `writeSector` and `AddTempSector` don't do any additional verification either.

persist/sqlite/sectors.go
```diff
@@ -239,106 +218,112 @@ WHERE ss.sector_root=$1 AND (EXISTS (SELECT 1 FROM contract_sector_roots csr WHE
...
+			volumeSectorIDs = append(volumeSectorIDs, c.VolumeSectorID.Int64)
+			volumeDeltas[c.VolumeID.Int64]--
+		}
+	}
 
-	rows, err := tx.Query(selectQuery, afterSectorID, encode(lastAccess), sqlSectorBatchSize)
-	if err != nil {
-		return nil, fmt.Errorf("failed to select volume sectors: %w", err)
+	if len(repairIDs) > 0 {
+		log.Warn("repairing sector reference counts", zap.Int("sectors", len(repairIDs)))
```
ChrisSchinnerl, line 247: This is just a safeguard right? We don't expect this to happen?

persist/sqlite/sectors.go
```diff
@@ -239,106 +218,112 @@ WHERE ss.sector_root=$1 AND (EXISTS (SELECT 1 FROM contract_sector_roots csr WHE
...
+	Referenced     bool
 }
 
-func updatePruneableVolumeSectors(tx *txn, lastAccess time.Time, afterSectorID int64) (refs []volumeSectorRef, err error) {
-	const selectQuery = `
-SELECT vs.id, vs.volume_id, vs.sector_id
-FROM volume_sectors vs
-INNER JOIN stored_sectors ss ON vs.sector_id=ss.id
-WHERE ss.id > $1
-	AND ss.last_access_timestamp < $2
```
ChrisSchinnerl, line 254: Just a heads up that we are losing the 5 minute grace period with this change.
Which protected against a race in `VolumeManager.ReadSector` between calling `SectorMetadata` and `readLocation`. Now the location could be updated to hold a different sector in between those calls which results in wrong data being returned without us knowing.

However that is insanely unlikely and might not be worth the fix.

persist/sqlite/volumes.go
```diff
@@ -454,10 +470,38 @@ func sectorDBID(tx *txn, root types.Hash256) (id int64, err error) {
...
+WHERE vs.volume_id=$1 AND vs.volume_index >= $2 LIMIT 1`, volumeID, minIndex).Scan(&index)
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to check volume sector locks: %w", err)
+	}
+	return fmt.Errorf("sector at volume %d index %d is locked", volumeID, index)
+}
+
+func lockVolumeSector(tx *txn, volumeSectorID int64) error {
```
ChrisSchinnerl, line 493: (nit) might be worth a comment that any operation relying on `volume_sector_locks` requires a write transaction. Maybe even mention that this relies on SQLite's aggressive locking. So that we don't just copy this when adding support for a database with more granular locking.

ChrisSchinnerl (dismissed):

persist/sqlite/volumes.go
```diff
@@ -454,10 +470,38 @@ func sectorDBID(tx *txn, root types.Hash256) (id int64, err error) {
...
 func insertSectorDBID(tx *txn, root types.Hash256) (id int64, err error) {
-	err = tx.QueryRow(`INSERT INTO stored_sectors (sector_root, last_access_timestamp) VALUES ($1, $2) ON CONFLICT (sector_root) DO UPDATE SET last_access_timestamp=EXCLUDED.last_access_timestamp RETURNING id`, encode(root), encode(time.Now())).Scan(&id)
+	err = tx.QueryRow(`INSERT INTO stored_sectors (sector_root) VALUES ($1) ON CONFLICT (sector_root) DO UPDATE SET sector_root=EXCLUDED.sector_root RETURNING id`, encode(root)).Scan(&id)
 	return
 }
 
+func checkVolumeSectorLocks(tx *txn, volumeID int64, minIndex uint64) error {
+	var index uint64
+	err := tx.QueryRow(`SELECT vs.volume_index FROM volume_sectors vs
+INNER JOIN volume_sector_locks l ON l.volume_sector_id=vs.id
```
peterjan: Should this be a CROSS JOIN driven from volume_sector_locks, since SQLite otherwise scans every sector in the volume and probes the lock table per row. I tested it on a local database but I might've messed up the database shape or something that makes this a non-issue.

persist/sqlite/init.sql
```diff
@@ -33,9 +33,10 @@ CREATE INDEX wallet_events_maturity_height ON wallet_events(maturity_height DESC
 CREATE TABLE stored_sectors (
 	id INTEGER PRIMARY KEY,
 	sector_root BLOB UNIQUE NOT NULL,
-	last_access_timestamp INTEGER NOT NULL
+	ref_count INTEGER NOT NULL DEFAULT 0
```
peterjan: Did you consider adding a `CHECK` here or getting rid of the `DEFAULT` and specify on every insert? Seems like that would be safer for not that much effort.

n8mgr: A `CHECK` sure, but why get rid of the default? It always starts at 0.

persist/sqlite/sectors.go
n8mgr: Correct. It shouldn't happen, but I didn't want to assume that the node could never end up in a bad state and leave the node stuck.

persist/sqlite/sectors.go
n8mgr: The fix is in the next PR. In the unlikely scenario, the read is retried once.

persist/sqlite/init.sql
peterjan: Migrations 🤷🏻‍♂️ could forget about the column and re-insert all rows at 0. Doesn't really matter though, check is plenty.

ChrisSchinnerl (approved):

PR #1001, opened 2026-09-09
Fix sector migration locking the database for long periods

n8mgr:
Splits sector migration into two parts, like regular sector writes, so that the database lock is no longer held while waiting for disk IO to complete.


persist/sqlite/volumes.go
```diff
@@ -244,84 +244,109 @@ func (s *Store) MigrateSectors(ctx context.Context, volumeID int64, startIndex u
...
+		}
 
+		if migrateErr := migrateFn(from, to); migrateErr != nil {
+			log.Error("failed to migrate sector", zap.Error(migrateErr), zap.Uint64("index", from.Index), zap.Stringer("root", from.Root))
+			failed++
+			if err = s.writeTransaction(release); err != nil {
+				err = fmt.Errorf("failed to release locations: %w", err)
+				return
+			}
+			continue
```
peterjan, line 307: Do we want to `jitterSleep` on `continue` here? Probably not too important but figured I'd raise it anyway.

peterjan (dismissed):

host/storage/storage.go
ChrisSchinnerl, line 728: Not introduced in this PR but this delete is missing a mutex.

ChrisSchinnerl (approved):
One unrelated thing but the rest looks good to me

peterjan (approved):

chris124567 (commented):
https://github.com/SiaFoundation/hostd/blob/b2c91202e3571cdf5313d4e301786d3d6cbeccef/persist/sqlite/volumes.go#L262

Preexisting but don't we need to wait for any inflight writes into the volume to finish before deciding we are done?

n8mgr:
> Preexisting but don't we need to wait for any inflight writes into the volume to finish before deciding we are done?

Volumes are marked read-only before migration starts. They shouldn't have any inflight writes. If they do happen to have one, Shrink or Remove volume should fail anyway.

chris124567 (approved):

PR #1009, opened 2026-09-15
Cleanup sector root cache

n8mgr:
Fixes a memory leak where the in-memory sector root cache kept the roots of every resolved, renewed, and rejected contract for the lifetime of the process.


persist/sqlite/contracts.go
```diff
@@ -522,9 +543,15 @@ LIMIT $3)`
 
 // updateResolvedV2Contract clears a contract and returns its ID
 func updateResolvedV2Contract(tx *txn, contractID types.FileContractID, renewedDBID int64) (dbID int64, err error) {
-	const clearQuery = `UPDATE contracts_v2 SET renewed_to=$1 WHERE contract_id=$2 RETURNING id;`
+	index, err := lastScannedIndex(tx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get last scanned index: %w", err)
+	}
+	const clearQuery = `UPDATE contracts_v2 SET renewed_to=$1, last_updated_height=$2, last_updated_block_id=$3 WHERE contract_id=$4 RETURNING id;`
```
ChrisSchinnerl: I think this change here is unnecessary. The renewed contract won't be considered resolved by `V2SectorRoots` and `ExpiredV2Contracts` until its `contract_status` was updated so imo it should be sufficient to only update `last_updated_height` and `last_updated_block_id` whenever `contract_status` gets updated in the consensus subscription code.

Same is true for the update in `reviseV2Contract`. Without those two the `last_updated_height` and `last_updated_block_id` simply describe the last time `contract_status` was updated. Which might warrant renaming those columns to reflect that.

At least the tests still pass for me when removing these 2 updates locally.

persist/sqlite/migrations.go
```diff
@@ -13,6 +13,26 @@ import (
...
+	}
+	_, err := tx.Exec(`
+ALTER TABLE contracts_v2 ADD COLUMN last_updated_height INTEGER NOT NULL DEFAULT 0;
+ALTER TABLE contracts_v2 ADD COLUMN last_updated_block_id BLOB NOT NULL DEFAULT x'0000000000000000000000000000000000000000000000000000000000000000';
+UPDATE contracts_v2 SET last_updated_height=resolution_height, last_updated_block_id=resolution_block_id WHERE resolution_height IS NOT NULL;
+CREATE INDEX contracts_v2_contract_status_last_updated_height ON contracts_v2(contract_status, last_updated_height);`)
+	if err != nil {
+		return fmt.Errorf("failed to add last updated columns: %w", err)
+	}
+	_, err = tx.Exec(`UPDATE contracts_v2 SET last_updated_height=$1, last_updated_block_id=$2 WHERE resolution_height IS NULL`, index.Height, encode(index.ID))
```
ChrisSchinnerl: (nit) any reason this isn't also done before creating the index to avoid updating the index after building it?

host/contracts/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+	rc.contractSectors[id] = slices.Clone(roots)
+}
+
+// ExpireContracts removes the cached sector roots of contracts that were
+// rejected or resolved before expireHeight.
+func (rc *rootsCache) ExpireContracts(height uint64) error {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
```
ChrisSchinnerl, line 51: Should be safe to not hold the lock when fetching the expired contracts from the database considering it's only called from a single thread. Even if that weren't true it should still be safe I think. A parallel call with a greater `height` would just prune a superset of contracts.

host/contracts/cache.go
```diff
@@ -0,0 +1,49 @@
...
+func (rc *rootsCache) UpdateSectorRoots(id types.FileContractID, roots []types.Hash256) {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+	rc.contractSectors[id] = slices.Clone(roots)
+}
+
+// ExpireContracts removes the cached sector roots of contracts that were
+// rejected or resolved before expireHeight.
+func (rc *rootsCache) ExpireContracts(height uint64) error {
+	rc.mu.Lock()
```
peterjan, line 50: No need to hold the lock while fetching expired contracts

```
diff --git a/host/contracts/cache.go b/host/contracts/cache.go
index 41889f7..17ae765 100644
--- a/host/contracts/cache.go
+++ b/host/contracts/cache.go
@@ -33,16 +33,25 @@ func (rc *rootsCache) UpdateSectorRoots(id types.FileContractID, roots []types.H
 // ExpireContracts removes the cached sector roots of contracts that were
 // rejected or resolved before expireHeight.
 func (rc *rootsCache) ExpireContracts(height uint64) error {
-	rc.mu.Lock()
-	defer rc.mu.Unlock()
-	if height > rc.lastExpiredHeight {
-		expired, err := rc.store.ExpiredV2Contracts(rc.lastExpiredHeight, height)
+	rc.mu.RLock()
+	lastExpiredHeight := rc.lastExpiredHeight
+	rc.mu.RUnlock()
+
+	// query the store without holding the lock so concurrent SectorRoots
+	// readers are not blocked while the transaction runs
+	var expired []types.FileContractID
+	if height > lastExpiredHeight {
+		var err error
+		expired, err = rc.store.ExpiredV2Contracts(lastExpiredHeight, height)
 		if err != nil {
 			return fmt.Errorf("failed to get expired contracts: %w", err)
 		}
-		for _, id := range expired {
-			delete(rc.contractSectors, id)
-		}
+	}
+
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+	for _, id := range expired {
+		delete(rc.contractSectors, id)
 	}
 	rc.lastExpiredHeight = height
 	return nil
```

peterjan (changes requested):

ChrisSchinnerl (approved):

peterjan (approved):

